PR #9, opened 2021-11-05
Deploy to dev environment with Terraform

gunndabad:
# Description

Adds initial Terraform resource definitions and amends the CI pipeline to deploy to the dev environment when PRs are ready for review.

Also adds docs around environment setup, secrets etc.

## Link to Trello Card

https://trello.com/c/PVd5L7Ne/65-terraform-the-api-application

### DIFF
--- a/.github/workflows/actions/deploy-environment/action.yml
+++ b/.github/workflows/actions/deploy-environment/action.yml
@@ -29,7 +29,7 @@ runs:
     - name: Get Key Vault name
       id: keyvault
       run: |
-        KEY_VAULT_NAME=$(cat $TFVARS | grep -E '^key_vault_name' | cut -d'=' -f2 | tr -d '"' | tr -d ' ')
+        KEY_VAULT_NAME=$(jq -r '.key_vault_name' $TFVARS)
 
         if [ -z "$KEY_VAULT_NAME" ]; then
           echo "::error ::Failed to extract key_vault_name from $TFVARS"
--- a/.github/workflows/actions/deploy-environment/action.yml
+++ b/.github/workflows/actions/deploy-environment/action.yml
@@ -1,0 +1,74 @@
+name: Deploy PAAS environment
+
+inputs:
+  environment_name:
+    description: 'The name of the environment'
+    required: true
+  docker_image:
+    description: 'The Docker image to deploy to the environment'
+    required: true
+  azure_credentials:
+    description: 'JSON object containing a service principal that can read from Azure Key Vault'
+    required: true
+  terraform_vars:
+    description: 'Path to the tfvars file for the environment'
+    required: true
+  terraform_backend_vars:
+    description: 'Path to the backend tfvars file for the environment'
+    required: true
+
+outputs:
+  environment_url:
+    description: 'The base URL for the deployed environment'
+    value: ${{ steps.terraform.outputs.url }}
+
+runs:
+  using: composite
+
+  steps:
+    - name: Get Key Vault name
+      id: keyvault
+      run: |
+        KEY_VAULT_NAME=$(cat $TFVARS | grep -E '^key_vault_name' | cut -d'=' -f2 | tr -d '"' | tr -d ' ')
+
+        if [ -z "$KEY_VAULT_NAME" ]; then
+          echo "::error ::Failed to extract key_vault_name from $TFVARS"
+          exit 1
+        fi
+
+        echo ::set-output name=name::$KEY_VAULT_NAME
+      shell: bash
+      env:
+        TFVARS: ${{ inputs.terraform_vars }}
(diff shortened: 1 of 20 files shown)


azure/azuredeploy.json
```diff
@@ -0,0 +1,103 @@
...
+    "storageAccountName": {
+      "type": "string"
+    },
+    "keyVaultReaderServicePrincipalId": {
+      "type": "string"
+    }
+  },
+  "variables": {
+    "location": "[resourceGroup().location]",
+    "tags": {
```
saliceti, line 20: Do we need to add tags here? I believe CIP automatically propagates tags from the resource group to the resources it contains

azure/azuredeploy.json
```diff
@@ -0,0 +1,103 @@
...
+      "Service Line": "Teaching Workforce",
+      "Service Offering": "Database of Qualified Teachers"
+    },
+    "tenantId": "9c7d9dd3-840c-4b3f-818e-552865082e16",
+    "tfstateContainerName": "dqtapi-tfstate",
+    "deliveryTeamGroupId": "ce5793a4-b822-4018-a65d-6c04e96918b2"
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
```
saliceti, line 35: Can we enable blob versioning for security?
I think it's via https://docs.microsoft.com/en-us/azure/storage/blobs/versioning-enable?tabs=template

azure/azuredeploy.json
```diff
@@ -0,0 +1,103 @@
...
+                "backup",
+                "restore"
+              ]
+            }
+          },
+          {
+            "tenantId": "[variables('tenantId')]",
+            "objectId": "[parameters('keyVaultReaderServicePrincipalId')]",
+            "permissions": {
+              "secrets": [
```
saliceti, line 83: Shouldn't we leave only readonly policies?

.github/workflows/build.yml
```diff
@@ -43,18 +47,68 @@ jobs:
     - name: Docker image tag
       id: image
       run: |
-        echo ::set-output name=tag::ghcr.io/$(echo $GITHUB_REPOSITORY | tr '[:upper:]' '[:lower:]'):sha-$(echo $GITHUB_SHA | cut -c -7)
+        echo ::set-output name=tag::$CONTAINER_REGISTRY/$(echo $GITHUB_REPOSITORY | tr '[:upper:]' '[:lower:]'):sha-$(echo $GITHUB_SHA | cut -c -7)
```
saliceti, line 50: Do you mind changing to full SHA as it's what has been used in all other docker images?
such as: https://github.com/DFE-Digital/apply-for-teacher-training/pkgs/container/apply-teacher-training

.github/workflows/build.yml
```diff
@@ -43,18 +47,68 @@ jobs:
...
-        push: ${{ github.ref == 'refs/heads/main' }}
+        push: true
         tags: ${{ steps.image.outputs.tag }}
+
+  validate_terraform:
+    name: Validate Terraform
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
saliceti, line 71: Can we use v2?

.github/workflows/actions/deploy-environment/action.yml
```diff
@@ -0,0 +1,74 @@
...
+    value: ${{ steps.terraform.outputs.url }}
+
+runs:
+  using: composite
+
+  steps:
+    - name: Get Key Vault name
+      id: keyvault
+      run: |
+        KEY_VAULT_NAME=$(cat $TFVARS | grep -E '^key_vault_name' | cut -d'=' -f2 | tr -d '"' | tr -d ' ')
```
saliceti, line 32: tfvars.json may also be used to store terraform variables so they can be read with `jq`
example: https://github.com/DFE-Digital/register-trainee-teachers/blob/main/.github/workflows/deploy.yml#L83

azure/azuredeploy.json
gunndabad: Done. That certainly makes the template simpler.

azure/azuredeploy.json
gunndabad: Amended to have the `get` and `list` permissions only.

.github/workflows/build.yml
gunndabad: Amended to use v2 throughout

.github/workflows/actions/deploy-environment/action.yml
gunndabad: Converted the file to JSON and updated this script to use `jq`.

saliceti (approved):

PR #31, opened 2021-12-03
Deploy to test environment from main branch

gunndabad:
# Description

Deploy to test environment from main branch.

Also updates the docs to add a missing step around service principal roles.

## Link to Trello Card

https://trello.com/c/PVd5L7Ne/65-deploy-api-application-to-paas-test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -152,3 +152,28 @@ jobs:
         allow_issue_writing: false
         fail_action: true
         cmd_options: '-z "-configFile /zap/wrk/options.prop"'
+
+  deploy_test:
+    name: Deploy to test environment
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -152,3 +152,28 @@ jobs:
         allow_issue_writing: false
         fail_action: true
         cmd_options: '-z "-configFile /zap/wrk/options.prop"'
+
+  deploy_test:
+    name: Deploy to test environment
+    needs: [build, validate_terraform]
```
merrilmathew, line 158: Do we need to include deploy_dev as dependency here?.

gunndabad: image

PR #76, opened 2022-01-12
Put TRN Request operation

gunndabad:
### Context

First pass on Put TRN Request for Register. Validation, caching etc. will come in later PRs.

https://trello.com/c/FJHFPAxn/155-register-api-create-put-request

Adds the `PUT /trn-requests/{requestId}` operation for creating or retrieving a TRN.

If the `requestId` already exists in our DB, lookup its corresponding `TeacherId` and return its `TRN`, if it has one.
Otherwise, create a new teacher record in CRM with the details in the request. If we suspect a duplicate, don't allocate a TRN but assign a task for the record to be manually reviewed.

### Guidance to review

Probably best reviewed commit-by-commit.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


src/DqtApi/V2/Handlers/GetOrCreateTrnRequestHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+                        CountryCode = request.Qualification.CountryCode,
+                        Subject = request.Qualification.Subject,
+                        Class = request.Qualification.Class.ConvertToClassDivision(),
+                        Date = request.Qualification.Date
+                    }
+                });
+
+                if (!createTeacherResult.Succeeded)
+                {
+                    throw new NotImplementedException();
```
gunndabad, line 98: A future PR will add validation and will catch these errors more gracefully.

src/DqtApi/V2/ApiModels/ClassDivision.cs
```diff
@@ -0,0 +1,87 @@
...
+            if (Enum.IsDefined(typeof(dfeta_classdivision), (int)input))
+            {
+                result = (dfeta_classdivision)input;
+                return true;
+            }
+            else
+            {
+                result = default;
+                return false;
+            }
```
defong, line 84:
```csharp
var inputAsInt = (int)input; 
return Enum.TryParse((input.ToString(), out dfeta_classdivision result) && Enum.IsDefined(typeof(dfeta_classdivision), inputAsInt);
```

src/DqtApi/V2/ApiModels/ClassDivision.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            if (!TryConvertToClassDivision(input, out var result))
+            {
+                throw new FormatException($"Unknown {nameof(ClassDivision)}: '{input}'.");
+            }
+
+            return result;
+        }
+
+        public static bool TryConvertToClassDivision(this ClassDivision input, out dfeta_classdivision result)
```
defong, line 73: Should this be private?

defong (approved):
Nothing major

src/DqtApi/V2/ApiModels/ClassDivision.cs
gunndabad: I've kept these `Try*` methods public as I expect our validation implementation (Coming Soon ™) to use these rather than using the methods that throw.

src/DqtApi/V2/ApiModels/ClassDivision.cs
gunndabad: Can't use `Parse(input.ToString())` here since not all of the string values of the source and destination enums match up exactly. I have moved the bulk of the logic into extension methods on Enum, however.

defong: Opps, I meant `inputAsInt.ToString()`

PR #84, opened 2022-01-25
Improve CRM integration tests performance

gunndabad:
### Context

Split apart from one of the new APIs for Register - https://trello.com/c/ngFfEsOx/16-register-api-create-put-request-for-qts

* Adds dependency injection support for tests
This allows us to have dependencies between fixtures, something that isn't possible with out-of-the-box xUnit. Specifically `CrmClientFixture` is now injected via DI and is singleton-scoped, meaning the CRM `ServiceClient` is now created and re-used across all tests. This is an enabler for the next change...

* Re-use some test data in a data-driven test
Lots of data permutations in `Given_details_that_do_not_match_existing_records_allocates_trn` previously meant that the same entity was getting created and cleaned up many times needlessly. This sets it up once, re-uses it for all test cases then cleans it up.

* Batch the cleanup requests to CRM.
Previously every cleaned up entity was done in its own API call, now they all go in a single call. We will likely need to batch this in future when the  number of entities goes over whatever the CRM API's batch request limit is.

* Run most tests in parallel
Apart from {Get|Create}Teacher-like operations we can safely run tests in parallel.

The perf improvement here isn't super impressive but it's a sizeable improvement for the forthcoming API addition.

### Guidance to review

Probably best reviewed commit-by-commit.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/tests/DqtApi.Tests/DataverseIntegration/CreateTeacherTests.cs
+++ b/tests/DqtApi.Tests/DataverseIntegration/CreateTeacherTests.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Threading.Tasks;
 using DqtApi.DataStore.Crm;
 using DqtApi.DataStore.Crm.Models;
 using Microsoft.PowerPlatform.Dataverse.Client;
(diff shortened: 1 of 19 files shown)


defong (approved):
Looks like the test has 30 secs knocked off :muscle:

PR #85, opened 2022-01-26
Set ITT outcome endpoint

gunndabad:
### Context

https://trello.com/c/ngFfEsOx/16-register-api-create-put-request-for-qts

Adds the 'Set ITT outcome' endpoint for Register. Given a TRN, an outcome and an UKPRN find the matching Person + QTS records and update their status.

Part of this adds support for error codes so our API consumers have a nicer way of catching errors without string matching.

### Guidance to review

Best reviewed commit-by-commit.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


docs/api-specs/v2.json
```diff
@@ -5,6 +5,100 @@
...
+      "put": {
+        "tags": [
+          "IttOutcome"
+        ],
+        "summary": "Sets ITT outcome for a teacher",
+        "parameters": [
+          {
+            "name": "trn",
+            "in": "path",
+            "description": "The TRN of the teacher to set ITT outcome for.",
```
dankmitchell, line 18: @benilovj should we go with `trainee` at this point, rather than `teacher`?

dankmitchell: hmm actually, in the context of DQT - every record is a teacher even though we know that's not correct so let's leave as is.

MrKevJoy (approved):

src/DqtApi/V2/Handlers/SetIttOutcomeHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+            _dataverseAdapter = dataverseAdapter;
+        }
+
+        public async Task<SetIttOutcomeResponse> Handle(SetIttOutcomeRequest request, CancellationToken cancellationToken)
+        {
+            var teachers = (await _dataverseAdapter.GetTeachersByTrn(request.Trn, activeOnly: true)).ToArray();
+
+            if (teachers.Length == 0)
+            {
+                throw new ErrorException(10001);  // Teacher with specified TRN not found
```
defong, line 29: It feels like the error numbers more like 

```csharp
ErrorCode.TeacherWithSpecifiedTRNNotFound
```

gunndabad: @defong I like that. I'll change them up.

PR #161, opened 2022-03-17
Add internal route

merrilmathew:
### Context

Add internal route for the api app and network policy so that Prometheus can receive metrics from the api app through internal route. This will enable aggregation of metrics over the number of api app instances running.

### Trello card

https://trello.com/c/G6763qSL

### DIFF
--- a/terraform/api-app.tf
+++ b/terraform/api-app.tf
@@ -14,6 +14,13 @@ resource "cloudfoundry_route" "api_public" {
   space    = data.cloudfoundry_space.space.id
 }
 
+resource "cloudfoundry_route" "api_internal" {
+  count    = local.configure_prometheus_network_policy
+  domain   = data.cloudfoundry_domain.internal.id
+  hostname = var.api_app_name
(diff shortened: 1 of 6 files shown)


merrilmathew:
[Add internal route and network policy](https://trello.com/c/G6763qSL/277-add-internal-route-and-network-policy)

terraform/preprod.tfvars.json
```diff
@@ -19,5 +19,6 @@
       "node_locations": ["UKINT", "UK1", "MAN1", "MAN5", "DUB2"],
       "confirmations": 2
     }
-  }
+  },
+  "prometheus_app": "prometheus-tra-monitoring-dev"
```
NickGraham101, line 23: Should this be `preprod` rather than `dev` or do you share a prometheus instance across non-prod environments?

merrilmathew: We only have dev and prod deployments of prometheus instance so yes shared.

saliceti (approved):

PR #165, opened 2022-03-21
stop passing dateonly to crm rather than datetime

MrKevJoy:
### Context

Immediate fix for passing dateonly to crm rather than a date time.

This fix stops the api from throwing a 500 when calling find.

### Guidance to review

### Checklist

-   [ ] Attach to Trello card
-   [x] Rebased master


src/DqtApi/DataStore/Crm/Models/FindTeachersQuery.cs
```diff
@@ -10,7 +10,7 @@ public class FindTeachersQuery
         public string LastName { get; set; }
         public string PreviousFirstName { get; set; }
         public string PreviousLastName { get; set; }
-        public DateOnly? DateOfBirth { get; set; }
+        public DateTime? DateOfBirth { get; set; }
```
gunndabad, line 13: This should be `DateOnly` - the conversion to `DateTime` should be inside the `DataverseAdapter` method.

gunndabad (approved):

PR #166, opened 2022-03-21
Fix to handle nulls in find teacher

MrKevJoy:
### Context
Handler was previously not able to handle nulls when there are no query string parameters, the dataverseadapter returned null early, but the handler was tryting to call .select on null. This change fixes this.

Change handler to handle nulls & add a missing test.

### Guidance to review

### Checklist

-   [ ] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/tests/DqtApi.Tests/V2/Operations/FindTeachersTests.cs
+++ b/tests/DqtApi.Tests/V2/Operations/FindTeachersTests.cs
@@ -1,9 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net.Http;
 using DqtApi.DataStore.Crm;
 using DqtApi.DataStore.Crm.Models;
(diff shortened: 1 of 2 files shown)


tests/DqtApi.Tests/V2/Operations/FindTeachersTests.cs
```diff
@@ -76,6 +77,30 @@ public async Task Given_search_returns_a_result_returns_expected_response()
...
+
+            // Act
+            var response = await HttpClient.SendAsync(request);
+
+            // Assert
+            await AssertEx.JsonResponseEquals(
+                response,
+                expected: new
+                {
+                    results = default(List<FindTeacherResult>)
```
gunndabad, line 98: I think we should be returning an empty collection here, not `null`

gunndabad (approved):

PR #174, opened 2022-03-22
Set correct status to null when teacher is withdrawn

MrKevJoy:
### Context

Update set itt outcome to clear teacherstatusid/earlyyearsstatusid conditionally.

Include a summary of the change.

### Guidance to review

### Checklist

-   [ ] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/tests/DqtApi.Tests/DataverseIntegration/SetIttOutcomeForTeacherTests.cs
+++ b/tests/DqtApi.Tests/DataverseIntegration/SetIttOutcomeForTeacherTests.cs
@@ -111,49 +111,59 @@ namespace DqtApi.Tests.DataverseIntegration
         }
 
         [Theory]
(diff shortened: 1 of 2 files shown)


tests/DqtApi.Tests/DataverseIntegration/SetIttOutcomeForTeacherTests.cs
```diff
@@ -215,7 +215,47 @@ public void SelectIttRecord_given_multiple_matching_itt_records_returns_Multiple
             Assert.Null(failedReason);
         }
 
-        // SelectQtsRecord
+        [Fact]
```
gunndabad, line 218: Would be good to have some negative tests here to prove that the field is *not* nulled out for a non-withdrawn status

gunndabad (approved):

PR #303, opened 2022-10-10
Don't filter out inactive subjects in Get Teacher endpoint

gunndabad:
### Context

Previous change to ignore all inactive data was too aggressive; we do want to return subjects even if they're inactive.

Return inactive subjects from Get Teacher endpoint

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


MrKevJoy (approved):

gunndabad:
[Notify clients when a user's details have changed](https://trello.com/c/v88Yc3x8/172-notify-clients-when-a-users-details-have-changed)

PR #330, opened 2022-11-14
Add API endpoints for accepting NPQ qualifications

gunndabad:
Adds single endpoint for updating & creating npq qualifications.

## `PUT/v2/npq-qualifications`

```json
{
  "completionDate": "2022-11-13",
  "qualificationType": "NPQEL|NPQH|NPQSL|NPQLT|NPQLTD|NPQLBC",
  "trn": "1234567"
}
```

All fields are mandatory except completionDate . If `completionDate` is in the future the endpoint will return an error.

If active teacher with specified TRN doesn't exist the endpoint will return a 404 response.

When completionDate is not null & an existing `dfeta_qualification`  matches the `qualification_type` then this record will be updated, otherwise a new `dfeta_qualification` will be created.

When completionDate is null & the teacher already has an awarded NPQ with the specified `qualification_type` then NPQ qualification will not be marked as not awarded.


cwrw (commented):
Hello @gunndabad, thanks for sharing the proposed API for accepting NPQ qualifications.

There were a few questions and suggestions that came up on a slack thread, which we can discuss tomorrow in the team catchup:
- A participant ID and qualification type can be used to uniquely distinguish a participant's outcome, thus the outcome ID is not necessary
- On our side an outcome will be immutable, meaning we can generate multiple outcome IDs per participant (if anything were to change) so we prefer not to send those across as they won't be unique
- We can use a participant ID and qualification type in both endpoints to create/update an NPQ qualification. (as suggested by @mattonfoot)
- What do we do with voided NPQ outcomes possibly made in error?
- In the API we are planning to surface to providers, we can allow them to create outcomes with the status failed then passed and keep updating as necessary, including the completion date. This makes deciding what to send to your API a bit more complex, and for which endpoint to hit (PUT vs POST). We will need to track if we already sent you a request.
However I wonder if we can simplify this by having one endpoint that's idempotent if an outcome exists, or updates the completion date in place otherwise? 

Thanks!

cwrw:
ah also I was wondering about bulk POSTing and if we will get the ability to send multiple outcomes at the same time?

MrKevJoy:
I've made the following changes to the request model as per todays meeting.

- Changed so that there is only one endpoint.
- Make completion date nullable.
- Remove outcome

cwrw:
Thanks for the update @MrKevJoy @gunndabad. We had a few questions/suggestions from our side:
> When completionDate is null & the teacher already has an awarded NPQ with the specified qualification_type then NPQ qualification will not be marked as not awarded.

I was hoping we could mark the NPQ qualification to show that it has now been voided as it could have been in error. Will we update anything else to indicate that this qualification was made in error otherwise?

> If active teacher with specified TRN doesn't exist the endpoint will return a 404 response.

Is it possible to return an explicit error/message indicating a TRN does not exist? As we will be capturing that on our side for flagging any records that need further investigation

> "qualificationType": "NPQEL|NPQH|NPQSL|NPQLT|NPQLTD|NPQLBC",

Could we also add the following types to the list: `NPQEYL` and `NPQLL` as those will be coming through later next year?

> Could we rely on TRN and qualificationType to identify a participant

Yes, we have double checked and all NPQ applicants have a TRN, and coupled with the qualification type can distinguish participant outcomes. However to ensure the TRNs we have are valid could we send you a list of all TRNs we expect to receive an outcome soon and you can check if they are valid? (can followup on slack) This will help us ensure the data we have is sound for now

cc @peteglondon

gunndabad:
@cwrw 
>I was hoping we could mark the NPQ qualification to show that it has now been voided as it could have been in error. Will we update anything else to indicate that this qualification was made in error otherwise?

I'll check with the business if they want us to mark these records as voided.

>Is it possible to return an explicit error/message indicating a TRN does not exist? As we will be capturing that on our side for flagging any records that need further investigation

There will be an error code in the JSON response for 'TRN doesn't exist' as well as the 404 status code.

>Could we also add the following types to the list: NPQEYL and NPQLL as those will be coming through later next year?

Sure.

We can do an upfront TRN check as well.

cwrw:
> I'll check with the business if they want us to mark these records as voided.
Maybe I misunderstood this in the meeting. I think we suggested that if an outcome was either sent by mistake and the participant had failed, or the outcome was sent accidentally, and needed to be voided (and could be sent again later) we could send an empty completion date to let you know and this would change the record internally for you, but not sure if that meant voided? Was it something else you'd mark the record with to hide/show it?

gunndabad:
> > I'll check with the business if they want us to mark these records as voided.
> 
> @gunndabad Maybe I misunderstood this in the meeting. I think we suggested that if an outcome was either sent by mistake and the participant had failed, or the outcome was sent accidentally, and needed to be voided (and could be sent again later) we could send an empty completion date to let you know and this would change the record internally for you, but not sure if that meant voided? Was it something else you'd mark the record with to hide/show it?

We would remove the completion date and remove the `awarded` flag which will prevent the record showing on the UI.

cwrw:
> We would remove the completion date and remove the awarded flag which will prevent the record showing on the UI.

Ah in that case yea that should be enough and no need to do any voiding. I'll send the TRN list on Monday, thanks!

DqtApi/src/DqtApi/DataStore/Crm/DataverseAdapter.SetNpqQualification.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        public async Task<SetNpqQualificationResult> SetNpqQualification(SetNpqQualificationCommand command)
+        {
+            var (r, _) = await SetNpqQualificationImpl(command);
+            return r;
+        }
+
+        internal async Task<(SetNpqQualificationResult Result, ExecuteTransactionRequest TransactionRequest)> SetNpqQualificationImpl(SetNpqQualificationCommand command)
+        {
+            var txnRequest = new ExecuteTransactionRequest()
```
gunndabad, line 20: Does we need a `ExecuteTransactionRequest` here? Looks like we're only ever having a single child request.

DqtApi/src/DqtApi/DataStore/Crm/SetNpqQualificationResult.cs
```diff
@@ -0,0 +1,28 @@
...
+            Succeeded = false,
+            FailedReasons = reasons
+        };
+    }
+
+    [Flags]
+    public enum SetNpqQualificationFailedReasons
+    {
+        None = 0,
+        MultipleNPQQualificationsWithQualificationType = 1
```
gunndabad, line 26:
```suggestion
        MultipleNpqQualificationsWithQualificationType = 1
```

DqtApi/src/DqtApi/Properties/StringResources.resx
```diff
@@ -195,4 +195,10 @@
   <data name="Errors.10018.Title" xml:space="preserve">
     <value>Husid registered for an existing teacher</value>
   </data>
+  <data name="Errors.10019.Title" xml:space="preserve">
+    <value>CompletionDate cannot be in the future</value>
+  </data>
+  <data name="Errors.10020.Title" xml:space="preserve">
+    <value>Multiple Qualifications with qualification type.</value>
```
gunndabad, line 202:
```suggestion
    <value>Multiple qualifications with qualification type.</value>
```

DqtApi/src/DqtApi/V2/Handlers/SetNpqQualificationHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+            else if (contacts.Length > 1)
+            {
+                throw new ErrorException(ErrorRegistry.MultipleTeachersFoundWithSpecifiedTrn());
+            }
+
+            var setNpqQualificationResult = await _dataverseAdapter.SetNpqQualification(new SetNpqQualificationCommand()
+            {
+                CompletionDate = request.CompletionDate.ToDateTime(),
+                //QualificationType = (dfeta_qualification_dfeta_Type)(request.QualificationType?.ConvertToQualificationType())
```
gunndabad, line 48: Don't we need this?

DqtApi/src/DqtApi/Validation/ErrorRegistry.cs
```diff
@@ -62,6 +62,8 @@ public static class ErrorRegistry
         public static Error TeacherAlreadyHasTsPersonIdDefined() => CreateError(10017);
 
         public static Error ExistingTeacherAlreadyHasHusId() => CreateError(10018);
+        public static Error QualificationCompletionDateInTheFuture() => CreateError(10019);
+        public static Error MultipleNPQQualificationWithQualificationType() => CreateError(10020);
```
gunndabad, line 66:
```suggestion
        public static Error MultipleNpqQualificationWithQualificationType() => CreateError(10020);
```

DqtApi/src/DqtApi/DataStore/Crm/SetNpqQualificationCommand.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using DqtApi.DataStore.Crm.Models;
+
+namespace DqtApi.DataStore.Crm
+{
+    public class SetNpqQualificationCommand
+    {
+        public DateTime? CompletionDate { get; set; }
+        public Guid TeacherId { get; set; }
+        public dfeta_qualification_dfeta_Type? QualificationType { get; set; }
```
gunndabad, line 10: Does this need to be nullable?

MrKevJoy (approved):

PR #393, opened 2023-02-17
Remove UseTrnGenerationApi feature flag

gunndabad:
Now that TRN Generation API is rolled out everywhere we no longer need to keep its use behind a feature flag.


hortha (approved):
Just wondering if we want to consciously leave the ability to use feature management (via nuget package and service registration) in this project even though it will not be used for the time being ?

PR #397, opened 2023-02-24
Add GET v3/certificates/qts endpoint

wilpow-github:
### Context

Users of the Access your teaching qualifications UI need to be able to download a QTS certificate. This ticket is for adding a new endpoint to the v3 API to allow them to do this.

Add a new endpoint at /v3/certificates/qts that returns a PDF of the QTS certificate with the user’s details in. The certificate should be printable only (not editable).

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


terraform/storage.tf
```diff
@@ -0,0 +1,24 @@
+resource "azurerm_storage_account" "qualified-teachers-api" {
```
wilpow-github, line 1: This could probably do with a close eye, as I've not been able to test

terraform/variables.tf
```diff
@@ -103,6 +103,10 @@ variable "hostnames" {
   type    = list(any)
 }
 
+variable "qualified_teachers_api_storage_account_name" {
```
wilpow-github, line 106: Haven't added this to the .tfvars files yet as not sure what name to use. Also there's already a var defined called storage_account_name (e.g. storage_account_name = "s165d01dqtapitfstatedv") that doesn't appear to be used anywhere

terraform/storage.tf
gunndabad, line 1:
```suggestion
resource "azurerm_storage_account" "app-storage" {
```

terraform/variables.tf
gunndabad: `storage_account_name` is used by this workflow action https://github.com/DFE-Digital/qualified-teachers-api/blob/main/.github/workflows/actions/deploy-environment/action.yml#L40

For naming, there's a convention that's enforced so something like `s165d01dqtapiappdv` should do it (the same as `storage_account_name` with `tfstate` swapped for `app`).

QualifiedTeachersApi/src/QualifiedTeachersApi/V3/Handlers/GetQtsCertificateHandler.cs
```diff
@@ -0,0 +1,131 @@
...
+            pdfData = stream.ToArray();
+        }
+
+        return new FileContentResult(pdfData, "application/pdf")
+        {
+            FileDownloadName = "QTS Certificate.pdf"
+        };
+    }
+
+    public async Task<FileResult> Handle(GetQtsCertificateRequest request, CancellationToken cancellationToken)
```
gunndabad, line 49: Could we return `Stream` here (or `byte[]`) instead of `FileResult`? Generally we try to keep these handlers decoupled from MVC types.

gunndabad (approved):

PR #434, opened 2023-03-29
Add new v3 GET teachers endpoint with TRN parameter

wilpow-github:
### Context

The Right to teach service needs to able to retrieve data for a teacher by a TRN. The existing v3 endpoints are authorised by an ID access token, which has a TRN baked in. We need an alternative endpoint where the TRN can be specified.

Add a `GET /v3/teachers/<trn>` endpoint. It should be authorised using the ApiKey authorization policy. It should return data in the same format as the existing /v3/teacher endpoint, for the TRN specified in the URL. If the TRN is not 7 digits, return a 400 response. If no teacher exists with the specified TRN, return a 404.

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/QualifiedTeachersApi/src/QualifiedTeachersApi/V3/Controllers/TeacherController.cs
+++ b/QualifiedTeachersApi/src/QualifiedTeachersApi/V3/Controllers/TeacherController.cs
@@ -12,7 +12,6 @@ using Swashbuckle.AspNetCore.Annotations;
 namespace QualifiedTeachersApi.V3.Controllers;
 
 [ApiController]
-[Route("teacher")]
 public class TeacherController : Controller
 {
     private readonly IMediator _mediator;
@@ -24,7 +23,7 @@ public class TeacherController : Controller
 
     [Authorize(AuthorizationPolicies.IdentityUserWithTrn)]
     [HttpGet]
-    [Route("")]
+    [Route("teacher")]
     [SwaggerOperation(
         summary: "Get teacher details",
         description: "Gets the details of the currently authenticated teacher")]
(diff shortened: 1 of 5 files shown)


QualifiedTeachersApi/src/QualifiedTeachersApi/V3/Requests/GetTeacherRequest.cs
```diff
@@ -1,9 +1,11 @@
-﻿using MediatR;
+﻿using System.ComponentModel.DataAnnotations;
+using MediatR;
 using QualifiedTeachersApi.V3.Responses;
 
 namespace QualifiedTeachersApi.V3.Requests;
 
 public record GetTeacherRequest : IRequest<GetTeacherResponse>
 {
+    [RegularExpression(@"^\d{7}$", ErrorMessage = "TRN must be 7 digits")]
```
gunndabad, line 9: Could we put this into `Validator` class?

gunndabad (approved):

QualifiedTeachersApi/src/QualifiedTeachersApi/V3/Controllers/TeacherController.cs
```diff
@@ -13,7 +13,6 @@ namespace QualifiedTeachersApi.V3.Controllers;
 
 [ApiController]
 [Route("teacher")]
```
gunndabad, line 15:
```suggestion
```

QualifiedTeachersApi/src/QualifiedTeachersApi/V3/Controllers/TeacherController.cs
```diff
@@ -23,12 +22,14 @@ public TeacherController(IMediator mediator)
         _mediator = mediator;
     }
 
+    [Authorize(AuthorizationPolicies.IdentityUserWithTrn)]
     [HttpGet]
     [Route("")]
```
gunndabad, line 27:
```suggestion
    [Route("teacher")]
```

PR #481, opened 2023-04-18
Add support for new stated first, middle & last name CRM fields

hortha:
### Context

When a new contact record is created via the DQT API, the FirstName provided is split such that the FirstName written to DQT is always a single word (any additional words are pre-pended to MiddleName). This normalisation is so that the matching process, which relies on exactly matching names, is more robust.

We need to ensure the name’s we’re setting on a Name change incident also have a single-word first name so that when the change is applied we don’t break matching processes. At the same time, we want to ‘round-trip’ names correctly (i.e. exactly as the user entered them) and store the names un-altered until such a time we can improve the matching processes to work with multi-word first names.

Changes aroung creating a teacher, changing teacher name + getting teacher information + associated tests

### Guidance to review

N/A

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Don't break DQT's matching when a multi-word first name change is submitted](https://trello.com/c/cYg80Okf/917-dont-break-dqts-matching-when-a-multi-word-first-name-change-is-submitted)

gunndabad (approved):

PR #489, opened 2023-04-24
Fix v3 swagger gen for Option<T> members

gunndabad:
Swashbuckle's built-in `ISchemaGenerator` can't handle `Option<T>` types correctly and it's unfortunately not easily extensible. This copies the in-box implementation an adds support for these types.

This is a quick fix; the long-term solution will use NSwag instead of Swashbuckle and we can look at a proper solution during migration.

### DIFF
--- a/QualifiedTeachersApi/src/QualifiedTeachersApi/Infrastructure/Swagger/SchemaGenerator.cs
+++ b/QualifiedTeachersApi/src/QualifiedTeachersApi/Infrastructure/Swagger/SchemaGenerator.cs
@@ -1,0 +1,503 @@
+﻿#nullable disable
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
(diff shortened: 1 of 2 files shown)


QualifiedTeachersApi/src/QualifiedTeachersApi/Infrastructure/Swagger/SchemaGenerator.cs
```diff
@@ -0,0 +1,503 @@
...
+        }
+
+        return schema;
+    }
+
+    private DataContract GetDataContractFor(Type modelType)
+    {
+        var effectiveType = modelType;
+
+        if (modelType.IsGenericType && modelType.GetGenericTypeDefinition() == typeof(Option<>))
```
gunndabad, line 174: This conditional is the only change over the standard built-in implementation.

MrKevJoy (approved):

PR #558, opened 2023-05-19
Add read-only mode

gunndabad:
When we move the DQT API from PAAS to AKS we want to do it without impacting the availability of services that depend on the API. The reason for the migration incurring down time is the database move; writes need to be stopped to the existing DB so that it can be copied and restored in the new DB without missing any data. Most of our endpoints don’t actually use the DB, they talk to CRM directly, so we don’t need to take these endpoints down.

This change adds a filter that checks the `ReadOnlyMode` value from configuration. If that value is `true`, requests to any endpoints that are not decorated with `SupportsReadOnlyMode` will return a `503` response.


gunndabad:
[DQT API 'read only' mode](https://trello.com/c/ynJ3kitE/1021-dqt-api-read-only-mode)

MrKevJoy (approved):

PR #560, opened 2023-05-20
Use NSwag for API spec generation

gunndabad:
We're planning on moving to FastEndpoints for API endpoints shortly and that uses NSwag instead of Swashbuckle for Swagger generation. In preparation for that this PR moves us to NSwag.

This is largely a port of the same customisations we had before with a few improvements: 
- I've added the security scheme for ID access tokens (which was missing before) and the appropriate logic to derive the scheme from the authorization policy. Swagger UI support is added too.
- 'Global' responses and security requirements have been added at the document level rather than being added for every operation (e.g. 429 response, the 'API key' security scheme).
- V3 endpoints are fleshed out a little more, with PDF-returning endpoints now correctly documenting their returned content type - `application/pdf`.

There's a lot of churn in the generated specs. I've checked over v3 and it all looks correct. Since we're no longer planning on adding to v1 and v2 I've left them largely as-is.


QualifiedTeachersApi/src/QualifiedTeachersApi/Infrastructure/OpenApi/HandleOptionJsonSchemaGenerator.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+    }
+
+    public override TSchemaType GenerateWithReferenceAndNullability<TSchemaType>(
+        ContextualType contextualType,
+        bool isNullable,
+        JsonSchemaResolver schemaResolver,
+        Action<TSchemaType, JsonSchema>? transformation = null)
+    {
+        if (contextualType.Type.IsGenericType && contextualType.Type.GetGenericTypeDefinition() == typeof(Option<>))
```
gunndabad, line 22: This makes `Option<T>` properties appear the same as `T` in the generated specs.

QualifiedTeachersApi/src/QualifiedTeachersApi/Infrastructure/OpenApi/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+            });
+        }
+
+        services.AddSingleton<IStartupFilter, OpenApiEndpointsStartupFilter>();
+
+        return services;
+    }
+}
+
+public class OpenApiEndpointsStartupFilter : IStartupFilter
```
hortha, line 70: Not in a separate file ?

hortha (approved):

QualifiedTeachersApi/src/QualifiedTeachersApi/Infrastructure/OpenApi/ServiceCollectionExtensions.cs
gunndabad: There's a lot of cross-over between the two types so I felt having them in a single file would be easier.

gunndabad: There's a follow-up PR coming where there's a lot less cross-over; I'll look to move it out into its own file with that change.

PR #567, opened 2023-05-22
Updated rate limiting to use .NET 7.0 + new Redis implementation

hortha:
### Context

ID has recently been moved to a new rate limiting library that sits on top of the .NET 7 rate limiting foundations. We should port that implementation to QT API so we’re using the same approach everywhere.

Port the Redis rate limiting implementation from ID into QT API. Note that this app does have client-specific rules in place currently and we should maintain that if possible.

### Guidance to review

Tested manually locally.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Port new rate limiting bits from ID into QT API](https://trello.com/c/NB3QxIw1/1026-port-new-rate-limiting-bits-from-id-into-qt-api)

gunndabad (approved):

PR #585, opened 2023-06-06
Reduce DqtReportingService metrics to the one we actually use

gunndabad:
I went a little overboard when adding metrics to the first iteration of the `DqtReportingService`. In reality, we only use one: total updates processed. This PR removes the fine grained metrics to emit only one - `updates processed`. We issue one of those for each entity type, instead of one per tick of the timer.

### DIFF
--- a/QualifiedTeachersApi/src/QualifiedTeachersApi/Services/DqtReporting/DqtReportingService.cs
+++ b/QualifiedTeachersApi/src/QualifiedTeachersApi/Services/DqtReporting/DqtReportingService.cs
@@ -1,5 +1,4 @@
 ﻿using System.Data;
-using System.Diagnostics;
 using System.ServiceModel;
 using System.Text;
 using Microsoft.ApplicationInsights;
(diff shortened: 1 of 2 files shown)


QualifiedTeachersApi/src/QualifiedTeachersApi/Services/DqtReporting/DqtReportingService.cs
```diff
@@ -148,44 +147,20 @@ internal async Task ProcessChanges(CancellationToken cancellationToken)
...
-                        });
 
                         lock (metricsGate)
                         {
                             totalUpdates += processedForEntityType;
                         }
                     }
+                    catch (Exception ex)
+                    {
+                        throw new Exception($"Failed processing changes for '{entityType}' entity.", ex);
```
gunndabad, line 162: Adding the entity type into the logged exception here so it's a little easier to debug issues.

hortha (approved):

PR #593, opened 2023-06-12
Add Terraform for deploying to AKS

gunndabad:
This PR moves the existing Terraform into a `paas` subfolder and creates a corresponding `aks` folder for setting up the application in AKS.

Since this repository is going to ultimately replace DQT and become TRS - 'Teaching Record System', I've taken this opportunity to change the names of the resources we're creating in AKS. Instead of `dqtapi` I've used `trs`.

There a few changes in the upstream `terraform-modules` repo that are not yet available within the `testing` tag so I've referenced `main` for the time being.


.github/workflows/actions/deploy-paas-environment/action.yml
```diff
@@ -47,16 +47,7 @@ runs:
       shell: bash
       env:
         TFVARS: ${{ inputs.terraform_vars }}
-      working-directory: terraform
-
-    - name: Extract version
```
gunndabad, line 52: This `Extract version` job was redundant so I've removed it.

.github/workflows/actions/deploy-paas-environment/action.yml
```diff
@@ -70,18 +61,16 @@ runs:
...
+        terraform_version: 1.5.0
         terraform_wrapper: false
 
     - name: Terraform
       id: terraform
       run: |
         make ci ${{ inputs.environment_name }} terraform-apply
-        cd terraform
-        echo "api_url=https://$(terraform output -raw api_fqdn)/" >> $GITHUB_OUTPUT
+        (cd terraform/paas && echo "api_url=https://$(terraform output -raw api_fqdn)/" >> $GITHUB_OUTPUT)
```
saliceti, line 71: Why the subshell with () ?
Also you can use terraform -chdir to avoid cd xxx

terraform/aks/databases.tf
```diff
@@ -0,0 +1,36 @@
+locals {
+  postgres_connection_string = "Server=${module.postgres.host};Database=${module.postgres.name};Port=${module.postgres.port};User Id=${module.postgres.username};Password='${module.postgres.password}';Ssl Mode=Require;Trust Server Certificate=true"
```
saliceti, line 2: Can we use the new module output?

gunndabad: The `url` output has a similar problem to Redis in that it uses a format that doesn't work for .NET. I could add a `connection_string` output to the postgres module similar to what we did for Redis?

saliceti: Yes let's do it now

saliceti (approved):

PR #676, opened 2023-07-25
Bugfix for slugid fallback

MrKevJoy:
### Context

This bugfix addresses problems with dead jobs that register received last night after they implemented the slugid changes. This PR introduces a fallback so that when slugid is passed and there isn't a match, it drops back to what it was doing previously before the slugid changes.

- updated tests.
- updated datavers method.

### Guidance to review

Include any useful information needed to review this change.
Include any dependencies that are required for this change.

### Checklist

-   [ ] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Dqt/DataverseAdapter.UpdateTeacher.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Dqt/DataverseAdapter.UpdateTeacher.cs
@@ -273,19 +273,27 @@ public partial class DataverseAdapter
                 Guid ittProviderId,
                 string slugId)
         {
-            // if SlugId is not passed in
-            // Find an active ITT record for the specified ITT Provider.
-            // otherwise find active ITT for the slugid (which should be one).
+            // if SlugId is passed in, use slugid for matching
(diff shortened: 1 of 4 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Dqt/DataverseAdapter.UpdateTeacher.cs
```diff
@@ -273,19 +273,27 @@ public UpdateTeacherHelper(DataverseAdapter dataverseAdapter, UpdateTeacherComma
...
+            {
+                activeForProvider = ittRecords
+                    .Where(r => r.dfeta_SlugId == slugId && r.StateCode == dfeta_initialteachertrainingState.Active)
+                    .ToArray();
+            }
+
+            if (activeForProvider.Length == 0)
+            {
+                activeForProvider = ittRecords
+                    .Where(r => r.dfeta_EstablishmentId?.Id == ittProviderId && r.StateCode == dfeta_initialteachertrainingState.Active)
```
gunndabad, line 294: I think we want a `&& string.IsNullOrEmpty(r.dfeta_SlugId)` too (to ensure we don't match on a different slug ID)

PR #712, opened 2023-08-10
Add an error page for users who aren't set up with any roles

gunndabad:


gunndabad:
[Add error screen for user with no roles](https://trello.com/c/vVWQSCsj/1164-add-error-screen-for-user-with-no-roles)

MrKevJoy (approved):

PR #754, opened 2023-08-16
Add TrnToken to TrnRequest and AYTQ link to response if it's populated

hortha:
### Context

Apply for QTS manage their own communications for users who’ve been awarded QTS. We want such users to be redirected to Access your teaching qualifications with a magic link so that they can shortcut the ID registration process, just as we do for non-overseas teachers.

Amend the GetOrCreateTrnRequest endpoint to add an AccessYourTeachingQualificationsLink property to the response. The link should generated when the QtsDate in the original request is non-`null` and the token stored in the trn_requests table (so it can be retrieved later). (The property should be an Option<string> so that it’s omitted completely in cases where we don’t generate a link.) The link should be generated in the same way as we do in the automated jobs (i.e. call the TRN token endpoint in ID and stitch the token into the Access your teaching qualifications endpoint). We should persist the TRN token in the DB table, rather than the complete link.

### Guidance to review

Reasonably simple change.

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master


hortha:
[Add magic link to TRN endpoint](https://trello.com/c/VGGFhMsc/1173-add-magic-link-to-trn-endpoint)

TeachingRecordSystem/src/TeachingRecordSystem.Api/V2/Handlers/GetOrCreateTrnRequestHandler.cs
```diff
@@ -139,12 +148,25 @@ public async Task<TrnRequestInfo> Handle(GetOrCreateTrnRequest request, Cancella
                 throw CreateValidationExceptionFromFailedReasons(createTeacherResult.FailedReasons);
             }
 
+            var trnTokenRequest = new CreateTrnTokenRequest
```
gunndabad, line 151: Nit: could we move this request inside the `if` block?

TeachingRecordSystem/src/TeachingRecordSystem.Api/V2/Handlers/GetOrCreateTrnRequestHandler.cs
```diff
@@ -164,7 +186,8 @@ public async Task<TrnRequestInfo> Handle(GetOrCreateTrnRequest request, Cancella
             Status = status,
             QtsDate = qtsDate,
             PotentialDuplicate = status == TrnRequestStatus.Pending,
-            SlugId = request.SlugId
+            SlugId = request.SlugId,
+            AccessYourTeachingQualificationsLink = trnToken is not null ? Option.Some($"{_accessYourQualificationsOptions.BaseAddress}/qualifications/start?trn_token={trnToken}") : default
```
gunndabad, line 190: Now we have `/qualifications/start` repeated in a few places I wonder whether we should put that onto `AccessYourQualificationsOptions` - maybe `public const string StartUrlPath = "/qualifications/start`

gunndabad (approved):

TeachingRecordSystem/src/TeachingRecordSystem.Api/V2/Handlers/GetOrCreateTrnRequestHandler.cs
hortha: Done

PR #770, opened 2023-08-22
added functionality to deactivate a user

ihsanahmed-tpx:
### Context

We need to be able to deactivate users who should no longer have access to the TRS Support Console.

- added changes to deactivate a user
- added relevant test

### Guidance to review

I've had to use the prefix `OnPost` as per the [MS Docs](https://learn.microsoft.com/en-gb/aspnet/core/razor-pages/?view=aspnetcore-7.0&tabs=visual-studio#multiple-handlers-per-page) for the new method (`OnPostDeactivate`). Also made use of `asp-page-handler` which will render as `formaction` in the markup

### Checklist

-   [#1180] Attach to Trello card

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserActivationEventChanges.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserActivationEventChanges.cs
@@ -1,7 +1,0 @@
-namespace TeachingRecordSystem.Core.Events;
-
-[Flags]
-public enum UserActivationEventChanges
-{
-    Deactivated = 0
-}
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserActivationEventChanges.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserActivationEventChanges.cs
@@ -1,0 +1,7 @@
+namespace TeachingRecordSystem.Core.Events;
+
+[Flags]
+public enum UserActivationEventChanges
+{
+    Deactivated = 0
+}
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserDeactivatedEvent.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserDeactivatedEvent.cs
@@ -1,0 +1,10 @@
+namespace TeachingRecordSystem.Core.Events;
+
+public record UserDeactivatedEvent : EventBase
+{
+    public required User User { get; init; }
+
+    public required Guid UpdatedByUserId { get; init; }
+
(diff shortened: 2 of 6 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserDeactivatedEvent.cs
```diff
@@ -0,0 +1,10 @@
+namespace TeachingRecordSystem.Core.Events;
+
+public record UserDeactivatedEvent : EventBase
+{
+    public required User User { get; init; }
+
+    public required Guid UpdatedByUserId { get; init; }
```
gunndabad, line 7:
```suggestion
    public required Guid DeactivatedByUserId { get; init; }
```

TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserActivationEventChanges.cs
```diff
@@ -0,0 +1,7 @@
+namespace TeachingRecordSystem.Core.Events;
+
+[Flags]
+public enum UserActivationEventChanges
```
gunndabad, line 4: Don't think we need this - there will only ever be one option.

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Users/EditUser.cshtml.cs
```diff
@@ -91,6 +94,36 @@ public async Task<IActionResult> OnPost()
 
     }
 
+    public async Task<IActionResult> OnPostDeactivate()
+    {
+        var user = await _dbContext.Users.SingleOrDefaultAsync(u => u.UserId == UserId);
```
gunndabad, line 99:
```suggestion
        var user = await _dbContext.Users.SingleAsync(u => u.UserId == UserId);
```

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Users/EditUser.cshtml.cs
```diff
@@ -91,6 +94,36 @@ public async Task<IActionResult> OnPost()
...
+
+        if (user is null)
```
gunndabad, line 101: Don't think we need this - the `OnPageHandlerExecutionAsync` will cover it.

TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/UserActivationEventChanges.cs
ihsanahmed-tpx: removed

gunndabad (approved):

PR #771, opened 2023-08-22
Add warning to Edit User page for users missing CRM account

gunndabad:


TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Services/AzureActiveDirectory/IUserService.cs
```diff
@@ -1,6 +1,6 @@
 namespace TeachingRecordSystem.SupportUi.Services.AzureActiveDirectory;
 
-public interface IUserService
+public interface IAadUserService
```
hortha, line 3: Filename is the old interface name still

hortha (approved):

PR #792, opened 2023-09-01
Add HighPotentialITT to v3 IttProgrammeType enum

gunndabad:


TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/ApiModels/IttProgrammeType.cs
```diff
@@ -75,7 +75,10 @@ public enum IttProgrammeType
     ProviderLedUndergrad = 389040022,
 
     [Description("International qualified teacher status")]
-    InternationalQualifiedTeacherStatus = 389040023
+    InternationalQualifiedTeacherStatus = 389040023,
+
+    [Description("High potential ITT")]
+    HighPotentialITT = 389040024
```
hortha, line 81: Should this be HighPotentialItt or do we explicitly want the capitals in this case?

gunndabad: Yes, it's to match the v2 model. Other abbreviations are cased this way in that enum too.

hortha (approved):

PR #793, opened 2023-09-01
Reject change of dob or name page

hortha:
### Context

We’re moving the view/approve/decline change of name & DOB processes into the new TRS Support Console.

Add the /cases/{id}/reject page following the designs in Figma.

The back link should go to /cases/{id}.

If the specified case does not exist, return a 404.

If the specified case’s state is not Active, return a 400.

Show a selection of radio buttons with rejection reasons, as per the designs. If Confirm is clicked without an option being selected, show an error ‘Select the reason for rejecting this change’.

When Confirm is clicked, set the case’s status to Rejected with an incident resolution status matching the selected rejected reason, and redirect to /cases with a success notification banner, as shown in the designs.

Change no longer required is a special case which should cause the case's status to be changed to Canceled

### Guidance to review

CRM queries + UI + tests

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Cases/EditCase/Reject.cshtml.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Cases/EditCase/Reject.cshtml.cs
@@ -56,11 +56,11 @@ public class RejectModel : PageModel
         (Incident Incident, dfeta_document[] Documents)? incidentAndDocuments = await GetIncidentAndDocuments();
 
         var requestStatus = "rejected";
-        var flashMessage = "The user's record has not been changed and they have been notified.";
+        var flashMessage = "The user’s record has not been changed and they have been notified.";
         if (RejectionReasonChoice!.Value == CaseRejectionReasonOption.ChangeNoLongerRequired)
         {
             requestStatus = "cancelled";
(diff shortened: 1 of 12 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Cases/EditCase/Reject.cshtml.cs
```diff
@@ -1,13 +1,93 @@
...
+    }
+
+    public async Task<IActionResult> OnPost()
+    {
+        if (!ModelState.IsValid)
+        {
+            return this.PageWithErrors();
+        }
+
+        (Incident Incident, dfeta_document[] Documents)? incidentAndDocuments = await GetIncidentAndDocuments();
```
hortha, line 56: The alternative to re-getting the incident in order to get the incident ID is probably to refactor the various pages to use incident ID in the URL etc so it is passed in the route .. sort of like the ticket number in the URL but I guess it doesn't really matter if it was the incident ID instead

hortha:
[Reject change of DOB/name page](https://trello.com/c/Sb14hc2Q/1192-reject-change-of-dob-name-page)

gunndabad (approved):

PR #890, opened 2023-11-02
Add draft TRN API spec

gunndabad:


aje54 (commented):
Should we tighten it up for early releases of the API and add nino? Currently around 30% coverage of active teachers (70%) with TPS data. May reduce support noise / unhappy path churn? (TPS would always have it - not sure about EWC)

gunndabad:
> Should we tighten it up for early releases of the API and add nino? Currently around 30% coverage of active teachers (70%) with TPS data. May reduce support noise / unhappy path churn? (TPS would always have it - not sure about EWC)

Good shout - I've added a `nationalInsuranceNumber` property.

aje54 (approved):

PR #1003, opened 2023-12-15
Add TRN API skeleton

gunndabad:
Adds just enough so we can get a complete API spec from the generated swagger doc.

[Design doc](docs/api-designs/trn.md)


aje54 (commented):
looks good

aje54 (approved):

PR #1011, opened 2023-12-19
Add MQ Provider to DB

gunndabad:
Also updates the MQ model for events to capture provider. We allow for specifying a DQT provider name in these events as the conversion from DQT MQ establishments to TRS MQ providers is lossy.


gunndabad:
[Add provider to TRS MQ data model](https://trello.com/c/HqbqAc3x/1517-add-provider-to-trs-mq-data-model)

hortha (approved):

PR #1018, opened 2023-12-19
Add option to DB migration command for specifying target migration

gunndabad:
PRs are deployed to the dev environment automatically. If that PR contains any migrations that are subsequently changed (or not merged) then the environment is stuck with the original changes applied, which possibly blocks newer changes from being applied successfully.

This change adds an option to the CLI command we use for deploying migrations to allow specifying a target migration. With this, we can SSH into a pod and run it to selectively roll back applied migrations. We could automate this but it happens sufficiently rarely that this is probably good enough for now.


aje54 (approved):
Looks good.

PR #1020, opened 2023-12-19
Only use journey state for mutable data

gunndabad:
Until now we've used journey state as a kind of cache for any data that the journey needs. I've changed this around so that only mutable data is stored.

The filter approach we started for 'personId' is now extended to cover MQs and this moves to using strongly-typed features and extension methods instead of `HttpContext.Items`.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Infrastructure/Filters/CheckPersonExistsFilter.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Infrastructure/Filters/CheckPersonExistsFilter.cs
@@ -5,16 +5,20 @@ using TeachingRecordSystem.Core.Dqt.Queries;
 
 namespace TeachingRecordSystem.SupportUi.Infrastructure.Filters;
 
-public class CheckPersonExistsFilter : IAsyncResourceFilter, IOrderedFilter
+/// <summary>
+/// Checks that a Person exists with the ID specified by the personId route value or query parameter.
+/// </summary>
+/// <remarks>
+/// <para>Returns a <see cref="StatusCodes.Status400BadRequest"/> response if the request is missing the personId route value.</para>
+/// <para>
+/// Returns a <see cref="StatusCodes.Status404NotFound"/> response if no person with the specified ID exists or if
+/// <paramref name="requireQts"/> is <c>true</c> and the person does not have QTS.
+/// </para>
+/// <para>Assigns the <see cref="CurrentPersonFeature"/> on success.</para>
+/// </remarks>
+public class CheckPersonExistsFilter(ICrmQueryDispatcher crmQueryDispatcher, bool requireQts = false) : IAsyncResourceFilter
 {
-    private readonly bool _requireQts;
-
-    public int Order => -200;
-
-    public CheckPersonExistsFilter(bool requireQts = false)
-    {
-        _requireQts = requireQts;
-    }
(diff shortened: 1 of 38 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Infrastructure/Filters/CheckPersonExistsFilter.cs
```diff
@@ -38,22 +41,30 @@ public async Task OnResourceExecutionAsync(ResourceExecutingContext context, Res
...
         }
 
-        context.HttpContext.Items["CurrentPersonDetail"] = person;
+        context.HttpContext.SetCurrentPersonInfoFeature(person);
 
         await next();
     }
 }
+
+public class CheckPersonExistsFilterFactory(bool requireQts = false) : IFilterFactory, IOrderedFilter
```
gunndabad, line 62: This factory means we can use constructor injection on the filter rather than `HttpContext.Services.GetRequiredService...`

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/PageContextFeatures.cs
```diff
@@ -0,0 +1,39 @@
...
+using TeachingRecordSystem.Core.Dqt.Models;
+
+namespace TeachingRecordSystem.SupportUi;
+
+public static class HttpContextExtensions
+{
+    public static CurrentPersonFeature GetCurrentPersonFeature(this HttpContext context) =>
+        context.Features.GetRequiredFeature<CurrentPersonFeature>();
+
+    public static void SetCurrentPersonInfoFeature(this HttpContext context, CurrentPersonFeature currentPersonInfo) =>
```
hortha, line 12: maybe this should not contain the word "Info" for consistency i.e. SetCurrentPersonFeature

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Program.cs
```diff
@@ -89,13 +90,28 @@
...
             {
-                model.Filters.Add(new CheckPersonExistsFilter());
+                model.Filters.Add(new CheckPersonExistsFilterFactory());
+            });
+
+        options.Conventions.AddFolderApplicationModelConvention(
+            "/Mqs/DeleteMq",
+            model =>
+            {
+                model.Filters.Add(new ServiceFilterAttribute<CheckMandatoryQualificationExistsFilter>());
```
hortha, line 107: is there a reason why this isn't done using the "factory" idea like CheckPersonExistsFilterFactory ?

gunndabad: We can't use `ServiceFilterAttribute` for `CheckPersonExistsFilter` because of its `requireQts` parameter.

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/PageContextFeatures.cs
gunndabad: Good catch, fixed.

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Program.cs
hortha: ah

hortha (approved):

PR #1067, opened 2024-01-12
Rename test fixture and base class

gunndabad:
Other test projects have `HostFixture` and `TestBase`; the API test project has `ApiFixture` and `ApiTestBase` instead. This fixes the inconsistency.


TeachingRecordSystem/tests/TeachingRecordSystem.Api.Tests/ApiFixture.cs
```diff
@@ -15,11 +15,11 @@
 
 namespace TeachingRecordSystem.Api.Tests;
 
-public class ApiFixture : WebApplicationFactory<Program>
+public class HostFixture : WebApplicationFactory<Program>
```
hortha, line 18: File still seems to be called ApiFixture

hortha (approved):

PR #1100, opened 2024-01-23
Add a Migrated event for MQ records

gunndabad:
We're adding an additional event at migration time for MQs so we have a place to record data changes (e.g. provider mappings). This adds the relevant logic for creating that event.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
@@ -53,15 +53,20 @@ public class TrsDataSyncHelper(
     public static MandatoryQualification MapMandatoryQualificationFromDqtQualification(
         dfeta_qualification qualification,
         IEnumerable<dfeta_mqestablishment> mqEstablishments,
-        IEnumerable<dfeta_specialism> mqSpecialisms)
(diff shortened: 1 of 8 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
```diff
@@ -1011,31 +1024,52 @@ EventBase MapImportedEvent(EntityVersionInfo<dfeta_qualification> snapshot)
...
             };
         }
 
-        Events.Models.MandatoryQualification GetEventMandatoryQualification(dfeta_qualification snapshot)
+        EventBase MapMigratedEvent(EntityVersionInfo<dfeta_qualification> snapshot)
+        {
+            var eventMandatoryQualification = GetEventMandatoryQualification(snapshot.Entity, applyMigrationMappings: true);
+
+            var changes = MandatoryQualificationMigratedEventChanges.None |
+                (eventMandatoryQualification.Provider?.Name != eventMandatoryQualification.Provider?.DqtMqEstablishmentName ? MandatoryQualificationMigratedEventChanges.Provider : 0);
```
hortha, line 1041: So anything with the exact same name will not be seen as a change then?

gunndabad: Yes, that’s the idea.

hortha (approved):

PR #1101, opened 2024-01-24
Person sync tweaks

gunndabad:
Until now we've not synced contact records without a TRN but we have some MQs linked to a contact record that doesn't have a TRN. This removes the filter to sync all contacts; we can always remove excess data later on.

I've also added {Created|Updated|Deleted}On fields to Person to be consistent with what we're doing elsewhere. Once we've done a full sync a subsequent change will make the former two non-nullable.


gunndabad:
[Production MQ import dry run](https://trello.com/c/th3k4qTv/1744-production-mq-import-dry-run)

MrKevJoy (approved):

PR #1109, opened 2024-01-25
Amended MQ UI to treat MQ with end date as Passed

hortha:
### Context

Existing Mandatory Qualifications in CRM will not have a status set so when trying to Edit them in TRS there would not be an existing status.

However if an MQ has an end date then it implicitly indicates that it is Passed.

Amend the view MQ and edit MQ status UI to show an MQ as Passed if dfeta_MQ_Status is null and dfeta_MQ_Date is not null.

### Guidance to review

Simplish change

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Treat existing MQs with End Date as Passed](https://trello.com/c/LByuGFrq/1747-treat-existing-mqs-with-end-date-as-passed)

gunndabad (approved):

PR #1115, opened 2024-01-29
Skeleton authentication scheme setup for Authorize project

gunndabad:
This sets up most of the structure for the two-phase authentication for Authorize access.

The OneLogin handler takes a fixed set of credentials but we will have a different set per client. Accordingly, we will have a OneLogin scheme for every one of our clients. (A future PR will wire this up.)

The OneLogin handler needs a 'sign in scheme' that can persist the AuthenticationTicket over multiple requests. Typically this is Cookies but here we're using FormFlow state instead. The thinking behind is that we want to support multiple concurrent sign ins for different clients and each of those needs its own state. If we used cookies it would mean a cookie-per-client, which could be a lot of cookies (and a messy Cookies page). This is the `FormFlowJourneySignInHandler` type and there only needs to be one of these.

The third handler type is the
`MatchToTeachingRecordAuthenticationHandler`. This is the scheme that actually gets called by code that wants to authenticate the user. It delegates to the appropriate OneLogin scheme to sign the user in to One Login, then redirects the user to the NINO page, which will begin the matching process. Eventually a teaching record may be found and authentication will be considered complete.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.AuthorizeAccessToATeacherRecord/Infrastructure/FormFlow/DummyCurrentUserIdProvider.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.AuthorizeAccessToATeacherRecord/Infrastructure/FormFlow/DummyCurrentUserIdProvider.cs
@@ -1,0 +1,8 @@
+using TeachingRecordSystem.SupportUi.Infrastructure.FormFlow;
+
+namespace TeachingRecordSystem.AuthorizeAccessToATeacherRecord.Infrastructure.FormFlow;
+
+public class DummyCurrentUserIdProvider : ICurrentUserIdProvider
+{
+    public string GetCurrentUserId() => Guid.Empty.ToString();
+}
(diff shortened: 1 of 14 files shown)


gunndabad:
[Productionise One Login spike](https://trello.com/c/Cq7vuQe8/1746-productionise-one-login-spike)

TeachingRecordSystem/src/TeachingRecordSystem.AuthorizeAccessToATeacherRecord/Infrastructure/FormFlow/DummyCurrentUserIdProvider.cs
```diff
@@ -0,0 +1,8 @@
+using TeachingRecordSystem.SupportUi.Infrastructure.FormFlow;
+
+namespace TeachingRecordSystem.AuthorizeAccessToATeacherRecord.Infrastructure.FormFlow;
+
+public class DummyCurrentUserIdProvider : ICurrentUserIdProvider
```
gunndabad, line 5: This is temporary, until we have a way of assigning a session ID cookie that we can use in place of a user ID.

hortha (approved):

PR #1122, opened 2024-01-31
Add set-kubelogin to database backup workflow

johnake:
### Context

Add set-kubelogin to database backup workflow, as this is required after adding rbac authentication

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master


.github/workflows/backup-db.yml
```diff
@@ -30,6 +30,10 @@ jobs:
         with:
           creds: ${{ secrets.AZURE_CREDENTIALS }}
 
+      - uses: DFE-Digital/github-actions/set-kubelogin-environment@master
```
saliceti, line 33: kubelogin is used by get-cluster-credentials so it must happen before it

saliceti:
The confirmation must be skipped in the workflow. please use the same implementation as in the template: https://github.com/DFE-Digital/teacher-services-cloud/blob/main/templates/new_service/Makefile#L39

.github/workflows/backup-db.yml
johnake: That's already above get-credentials? I just removed the duplicate before it

Makefile
```diff
@@ -130,7 +130,7 @@ ci:	## Run in automation environment
 	$(eval DISABLE_PASSCODE=true)
 	$(eval AUTO_APPROVE=-auto-approve)
 	$(eval SP_AUTH=true)
-	$(eval CONFIRM_DEPLOY=true)
+	$(eval SKIP_CONFIRM=true)
```
saliceti, line 133: This is only half implemented. Please check the template to see how SKIP_CONFIRM is used

Makefile
```diff
@@ -128,9 +128,11 @@ set-azure-pd-subscription:
 
 ci:	## Run in automation environment
 	$(eval DISABLE_PASSCODE=true)
-	$(eval AUTO_APPROVE=-auto-approve)
 	$(eval SP_AUTH=true)
 	$(eval CONFIRM_DEPLOY=true)
+	$(eval AUTO_APPROVE=-auto-approve)
+	$(eval SKIP_AZURE_LOGIN=true)
```
saliceti, line 134: This is not used, please remove

saliceti (approved):

PR #1124, opened 2024-01-31
Add a debug page for overriding user information

gunndabad:
We need to be able to override the identity information we get from One Login for testing in non-production environments. This adds a debug page into the start of the journey where we can override that information and detach a `Person`, if one is already assigned.


gunndabad:
[Override One Login user info](https://trello.com/c/QaRGKkjl/1766-override-one-login-user-info)

MrKevJoy (approved):

PR #1130, opened 2024-02-02
Added person search service against TRS DB

hortha:
### Context

There are several places where we do a person search against DQT. Each of these searches has different matching criteria. We’ve been constrained by CRM on how smart we can make these searches since we don’t have access to the underlying DB. Now that we’re syncing person data to the TRS DB, we can do better.

This ticket is to begin a new search service that can be used with Authorize access only (initially).

Create a PersonSearchService with the following signature:

class PersonSearchService
{
  Task<IReadOnlyCollection<PersonSearchResult>> Search(IEnumerable<string[]> name, IEnumerable<DateOnly> dateOfBirth, string? nino, string? trn);
}
Note that the name and dateOfBirth parameter take collections (since we can get multiple names and/or DOBs from One Login).

The method should return all Person records that match on a first name + last name, a DOB then at least one of NINO or TRN.

The service should search the TRS DB, not DQT.

Name matching should allow for aliases.

For this iteration searching on previous names is out of scope, though the implementation should allow for adding this later.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Migrations/20240201153254_PersonSearchAttributeAndNameSynonyms.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Migrations/20240201153254_PersonSearchAttributeAndNameSynonyms.cs
@@ -1,0 +1,267 @@
+using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
+
+#nullable disable
+
+namespace TeachingRecordSystem.Core.DataStore.Postgres.Migrations
+{
+    /// <inheritdoc />
+    public partial class PersonSearchAttributeAndNameSynonyms : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
+                name: "name_synonyms",
+                columns: table => new
+                {
+                    name_synonyms_id = table.Column<long>(type: "bigint", nullable: false)
+                        .Annotation("Npgsql:ValueGenerationStrategy", NpgsqlValueGenerationStrategy.IdentityByDefaultColumn),
+                    name = table.Column<string>(type: "character varying(100)", maxLength: 100, nullable: false, collation: "case_insensitive"),
+                    synonyms = table.Column<string[]>(type: "text[]", nullable: false, collation: "case_insensitive")
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("pk_name_synonyms", x => x.name_synonyms_id);
+                });
+
+            migrationBuilder.CreateTable(
+                name: "person_search_attributes",
+                columns: table => new
+                {
+                    person_search_attribute_id = table.Column<long>(type: "bigint", nullable: false)
+                        .Annotation("Npgsql:ValueGenerationStrategy", NpgsqlValueGenerationStrategy.IdentityByDefaultColumn),
+                    person_id = table.Column<Guid>(type: "uuid", nullable: false),
+                    attribute_type = table.Column<string>(type: "character varying(50)", maxLength: 50, nullable: false, collation: "case_insensitive"),
+                    attribute_value = table.Column<string>(type: "character varying(100)", maxLength: 100, nullable: false, collation: "case_insensitive"),
+                    tags = table.Column<string[]>(type: "text[]", nullable: false),
+                    attribute_key = table.Column<string>(type: "character varying(50)", maxLength: 50, nullable: true, collation: "case_insensitive")
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("pk_person_search_attributes", x => x.person_search_attribute_id);
+                });
+
+            migrationBuilder.CreateIndex(
+                name: "ix_name_synonyms_name",
+                table: "name_synonyms",
+                column: "name",
+                unique: true);
+
+            migrationBuilder.CreateIndex(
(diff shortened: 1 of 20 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/PopulateAllPersonsSearchAttributesJob.cs
```diff
@@ -0,0 +1,18 @@
...
+using TeachingRecordSystem.Core.Jobs.Scheduling;
+
+namespace TeachingRecordSystem.Core.Jobs;
+
+[AutomaticRetry(Attempts = 0)]
+public class PopulateAllPersonsSearchAttributesJob(TrsDbContext dbContext, IBackgroundJobScheduler backgroundJobScheduler)
+{
+    public async Task Execute()
+    {
+        foreach (var person in dbContext.Persons.AsNoTracking())
```
gunndabad, line 13: I'm not sure exactly how this works internally but I suspect this may not work with the data volume we have on production. It may be better to get an `IAsyncEnumerable` and also only select the attribute(s) we care about; we certainly don't want to be pulling every `Person` record into memory at one time.
```
await foreach (var personId in dbContext.Persons.Select(p => p.PersonId).AsAsyncEnumerable())
{
    await backgroundJobScheduler.Enqueue<PopulatePersonSearchAttributesJob>(j => j.Execute(personId));
}
```

TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/PopulateNameSynonymsJob.cs
```diff
@@ -0,0 +1,69 @@
...
+using TeachingRecordSystem.Core.DataStore.Postgres;
+using TeachingRecordSystem.Core.DataStore.Postgres.Models;
+
+namespace TeachingRecordSystem.Core.Jobs;
+
+public class PopulateNameSynonymsJob(TrsDbContext dbContext)
+{
+    public async Task Execute(CancellationToken cancellationToken)
+    {
+        var httpClient = new HttpClient();
```
gunndabad, line 11: Can we inject this?

TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/PopulateAllPersonsSearchAttributesJob.cs
hortha: Funnily enough I did have it as IAsyncEnumerable originally but didn't think it would make it any difference to how it would buffer the results in memory i.e. as long as not adding ToListAsync() then it should NOT hold all records in memory even without IAsyncEnumerable

TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/PopulatePersonSearchAttributesJob.cs
```diff
@@ -0,0 +1,90 @@
...
+using TeachingRecordSystem.Core.DataStore.Postgres.Models;
+
+namespace TeachingRecordSystem.Core.Jobs;
+
+[AutomaticRetry(Attempts = 0)]
+public class PopulatePersonSearchAttributesJob(TrsDbContext dbContext)
+{
+    public async Task Execute(Guid personId)
+    {
+        var person = await dbContext.Persons.AsNoTracking().SingleAsync(p => p.PersonId == personId);
```
gunndabad, line 13: Can we call the SQL function that the trigger uses here instead of duplicating logic?

TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/PopulateAllPersonsSearchAttributesJob.cs
gunndabad: If that's the case then existing enumerable is ok. I still think it's worth the `Select(p => p.PersonId)`, though.

gunndabad (approved):

TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Migrations/20240201153254_PersonSearchAttributeAndNameSynonyms.cs
```diff
@@ -0,0 +1,267 @@
...
+    IN p_person_id uuid,
+    IN p_first_name character varying(100),
+    IN p_last_name character varying(100),
+    IN p_date_of_birth date,
+    IN p_national_insurance_number character(9),
+    IN p_trn character(7))
+LANGUAGE 'plpgsql'
+AS $BODY$
+BEGIN
+    DELETE FROM
```
gunndabad, line 160: Does this run in the same transaction as the insert? If a search occurs at the same time this is run do we still get the expected results?

TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Migrations/20240201153254_PersonSearchAttributeAndNameSynonyms.cs
```diff
@@ -0,0 +1,267 @@
...
+            var refreshNameProcedureSql = @"
+CREATE OR REPLACE PROCEDURE public.p_refresh_name_person_search_attributes(
+    IN p_person_id uuid,
+    IN p_first_name character varying,
+    IN p_last_name character varying,
+    IN p_attribute_key character varying)
+LANGUAGE 'plpgsql'
+AS $BODY$
+BEGIN
+    DELETE FROM
```
gunndabad, line 71: As the other; does this run in the same transaction etc.?

PR #1162, opened 2024-02-12
Store redirect URIs for One Login clients

gunndabad:
Adds redirect URI, post logout redirect URI and an authentication scheme name to `ApplicationUser` for One Login clients. Also adds an explicit `IsOidcClient` flag.

I've also hidden the private key by default (it's visible by clicking a 'Show key' checkbox).


gunndabad:
[Add authentication scheme information to DB](https://trello.com/c/Xp2ODbOz/1805-add-authentication-scheme-information-to-db)

PR #1165, opened 2024-02-13
Fix more MQ journey snags

gunndabad:
This makes the generated error summary appear above the heading and caption.

Also removes some MQ specialisms from the Add MQ journey; we don't want new records creating with legacy specialisms.


gunndabad:
[MQ snags](https://trello.com/c/pzGJGP4B/1816-mq-snags)

hortha (approved):

PR #1202, opened 2024-02-19
Remove unnecessary QTS assignment for adding an MQ in tests

gunndabad:
I amended the signature of `WithQts()` too; we don't use it here any more but it'll be useful in future.


gunndabad:
[MQ snagging (round 2)](https://trello.com/c/mZYVWvJa/1830-mq-snagging-round-2)

MrKevJoy (approved):

PR #1265, opened 2024-04-04
Added support for multiple establishment sources

hortha:
### Context

The TPS extract for workforce data contains Local Authority Code and Establishment Number combinations which do not map to GIAS establishments but DO map to other employers held internally in TPS tables.

Amend the existing TRS database schema to have an establishment source associated with an establishment.


hortha:
[Accommodate multiple establishment sources](https://trello.com/c/wD923jv6/1971-accommodate-multiple-establishment-sources)

gunndabad (approved):

PR #1292, opened 2024-04-22
Support trn_token in sign in journeys

gunndabad:
We want to support 'magic links' that we've emailed to people inviting them to AYTQ. This adds support for the `trn_token` parameter, as with the current ID solution. The token is associated with a TRN. If the verified last name and DOB match those on the teaching record for the token's TRN we automatically attach the user to the teaching record. If there's a discrepancy we fall back to the regular matching journey.

For now we're reaching directly into the ID database; normally we'd create APIs but this is temporary until we move all the tokens over from ID (at which point we'll likely need direct DB access any way). For testing this uses the EF in memory provider (setting up real DBs for testing this isn't worth it).

Much of this PR is refactoring to allow the Debug page to share some code in `SignInJourneyHelper`; `TryApplyTrnToken` is the meat of the changes.


gunndabad:
[Support magic links in Teacher Auth](https://trello.com/c/t6yqPzgv/105-support-magic-links-in-teacher-auth)

hortha (approved):

PR #1297, opened 2024-04-25
Create a support task when automatic matching fails

gunndabad:
When automatic matching fails a support task needs to be raised for support to find the record and associate it with the One Login User. This adds the DB schema for support tasks and wires up the journey to create one when a user is verified but the matching fails.


gunndabad:
[Create support task when a user could not be matched](https://trello.com/c/WYsqadzW/100-create-support-task-when-a-user-could-not-be-matched)

hortha (approved):

PR #1298, opened 2024-04-26
Change PersonSearchService to PersonMatchingService

gunndabad:
The policy used by the `Search` method on `PersonSearchService` is specific to matching in `AuthorizeAccess`; this moves the service into the `AuthorizeAccess` project accordingly and renames it to `PersonMatchingService`. There will be additional methods added to it shortly for returning suggested matches for the support back end.

I've also amended the implementation to use raw SQL; the resulting query is simpler than the previous EF-generated query and is about 2x faster from an initial test locally. (I expect it to be even quicker when run in production with real data.)


gunndabad:
[Resolve One Login 'not matched' support ticket](https://trello.com/c/slLElIRC/103-resolve-one-login-not-matched-support-ticket)

TeachingRecordSystem/src/TeachingRecordSystem.AuthorizeAccess/SignInJourneyHelper.cs
```diff
@@ -6,17 +6,17 @@
...
-using TeachingRecordSystem.Core.Services.PersonSearch;
 using TeachingRecordSystem.FormFlow;
 using TeachingRecordSystem.FormFlow.State;
 
 namespace TeachingRecordSystem.AuthorizeAccess;
 
 public class SignInJourneyHelper(
     TrsDbContext dbContext,
-    IPersonSearchService personSearchService,
+    IPersonMatchingService personSearchService,
```
hortha, line 19: maybe parameter should be called personMatchingService now?

TeachingRecordSystem/tests/TeachingRecordSystem.AuthorizeAccess.Tests/SignInJourneyHelperTests.cs
```diff
@@ -703,11 +609,11 @@ public class SignInJourneyHelperTests(HostFixture hostFixture) : TestBase(hostFi
             Assert.Equal(person.Trn, state.AuthenticationTicket.Principal.FindFirstValue(ClaimTypes.Trn));
         });
 
-    private SignInJourneyHelper CreateHelper(TrsDbContext dbContext, IPersonSearchService? personSearchService = null)
+    private SignInJourneyHelper CreateHelper(TrsDbContext dbContext, IPersonMatchingService? personSearchService = null)
```
hortha, line 612: Again maybe parameter should be called personMatchingService ?

hortha (approved):

PR #1305, opened 2024-05-07
Workforce data extract support ageing and non gias establishments

hortha:
### Context

Since the original job to import TPS extract data was written, we have discovered that the end date field is actually the "last known employed date".
To deduce an actual end date we need to use a combination of this date and the last extract date to see if this has stayed the same for over x months as an indication that this teacher is no longer working at this employed.
Also, now that we have imported establishments based on a spreadsheet + some range mappings provided by TPS, there is no longer the need to have certain fields be nullable.

DB changes, changes to the import process + an adhoc job to age the data (this maybe be removed at a later date if we don't believe the data is reliable enough) + test changes.

### Guidance to review

Not a huge amount of code changes in the end

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/ProcessEndedEmploymentsJob.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/ProcessEndedEmploymentsJob.cs
@@ -1,4 +1,4 @@
-﻿using TeachingRecordSystem.Core.Jobs.Scheduling;
+using TeachingRecordSystem.Core.Jobs.Scheduling;
 using TeachingRecordSystem.Core.Services.WorkforceData;
 
 namespace TeachingRecordSystem.Core.Jobs;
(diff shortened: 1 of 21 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/ProcessEndedEmploymentsJob.cs
```diff
@@ -0,0 +1,12 @@
+using TeachingRecordSystem.Core.Jobs.Scheduling;
+using TeachingRecordSystem.Core.Services.WorkforceData;
+
+namespace TeachingRecordSystem.Core.Jobs;
+
+public class ProcessEndedEmploymentsJob(IBackgroundJobScheduler backgroundJobScheduler)
+{
+    public async Task ExecuteAsync(CancellationToken cancellationToken)
+    {
+        await backgroundJobScheduler.Enqueue<TpsCsvExtractProcessor>(j => j.ProcessEndedEmployments(cancellationToken));
```
gunndabad, line 10: Is there any need to push this into its own job? Could we just process here instead?

hortha:
[Workforce data extract support ageing and non GIAS establishments](https://trello.com/c/85fTqEMB/127-workforce-data-extract-support-ageing-and-non-gias-establishments)

PR #1306, opened 2024-05-08
Add support for downloading evidence files from Blob Storage

gunndabad:
AYTQ cannot easily give us a blob URI with a SAS token. This adds support for retrieving evidence files using an authenticated `BlobClient` for pre-configured blob storage accounts. It uses a `DelegatingHandler` to intercept the request, check if it matches a known blob storage account URI then uses `BlobClient` (with a configured token) to download the file and create an `HttpResponseMessage`.


gunndabad:
[Configure blob storage key for evidence file downloads](https://trello.com/c/cvTBAhTH/138-configure-blob-storage-key-for-evidence-file-downloads)

MrKevJoy (approved):

PR #1318, opened 2024-05-13
Fix out of memory issue for workforce data import

hortha:
### Context

While running the job to import the 2nd months workforce data System.OutOfMemoryException was thrown.

Initial investigation has lead us to believe this is related the EF change tracking on a large number of entities (specifically all the Events that get raised as part of the processing).

Amended the SQL to update as much as possible in each call to the database.
Added transactions and improved events generation (removed need for EF change tracking).

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/NpgsqlExtensions.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/NpgsqlExtensions.cs
@@ -1,0 +1,117 @@
+using System.Text.Json;
+using Npgsql;
+using NpgsqlTypes;
+
+namespace TeachingRecordSystem.Core.DataStore.Postgres;
+
+public static class NpgsqlExtensions
+{
+    public static void WriteValueOrNull<T>(this NpgsqlBinaryImporter writer, T? value, NpgsqlDbType dbType)
+        where T : struct
+    {
+        if (value.HasValue)
+        {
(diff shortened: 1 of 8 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Extensions/NpgsqlExtensions.cs
```diff
@@ -0,0 +1,117 @@
+using System.Text.Json;
+using Npgsql;
+using NpgsqlTypes;
+
+namespace TeachingRecordSystem.Core.DataStore.Postgres.Extensions;
```
gunndabad, line 5: Can we move this up a level (saves an extra `using` to have the methods visible)?

TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Extensions/NpgsqlExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+
+public static class NpgsqlExtensions
+{
+    public static void WriteValueOrNull<T>(this NpgsqlBinaryImporter writer, T? value, NpgsqlDbType dbType)
+    where T : struct
```
gunndabad, line 10: Nit
```suggestion
        where T : struct
```

PR #1350, opened 2024-06-03
Consider first name synonyms when checking for existing contact

gunndabad:
This extends first name synonym application to the duplicate detection used in the v3 TRN request endpoint.

As part of this I've moved the synonym download code out of the existing job (that populates the TRS DB) into its own service.


gunndabad:
[Add synonym name matching support for V3 TRN API](https://trello.com/c/EvVUaLPB/182-add-synonym-name-matching-support-for-v3-trn-api)

MrKevJoy (approved):

PR #1351, opened 2024-06-04
Use email address when checking for an existing contact

gunndabad:
This amends the v3 TRN request endpoint to use email address as one of the matching fields. It also amends the request to support passing multiple email addresses, though only the first will be written to the `contact` record.


gunndabad:
[Use email address when searching for existing TRN holder](https://trello.com/c/T1wMs9WA/185-use-email-address-when-searching-for-existing-trn-holder)

MrKevJoy (approved):

PR #1361, opened 2024-06-10
Match on email address alone in duplicates query

gunndabad:
This amends the query used for finding potential duplicate contacts such that matching on email address alone is considered a match (previously at least two other attributes would need to be matched too).


gunndabad:
[Match on email field](https://trello.com/c/wyKe9IdH/192-match-on-email-field)

MrKevJoy (approved):

PR #1374, opened 2024-06-13
Add NINO and postcode from workforce data into person_search_attributes

hortha:
### Context

The current person search used by the One Login integration uses the NINO from DQT/TRS only. Now that we have NINO from workforce data to we should include this to improve the chances of matching using NINO. We may use the person’s postcode to match on in future as well so we should populate this too.

Amend the process that imports workforce data to populate person_search_attributes with the NINO and person’s postcode, where available.

N.B. a database trigger on the appropriate workforce data tables is probably the way to go for this, similar to what we’re doing with the person table.

### Guidance to review

DB migrations + code + tests

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Add NINO and postcode from workforce data into person_search_attributes](https://trello.com/c/hD9znNs0/51-add-nino-and-postcode-from-workforce-data-into-personsearchattributes)

gunndabad (approved):

PR #1383, opened 2024-06-18
Updated triggers to use IS DISTINCT FROM when comparing values which could be NULL

hortha:
### Context

Triggers on `person_employments` table were not updating `person_search_attributes` for some reason.

Investigation lead to finding that using `!=` in a SQL WHERE clause ignored records where either of the field values being compared were `NULL`.
This has been replaced by `IS DISTINCT FROM` in the `persons` and `person_employments` trigger functions which update the `person_search_attributes` table.

### Guidance to review

Database changes only.

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master


hortha:
[Fix issues with populating workforce data NINO and postcode in the person_search_attributes table](https://trello.com/c/ME5FdJWr/211-fix-issues-with-populating-workforce-data-nino-and-postcode-in-the-personsearchattributes-table)

PR #1391, opened 2024-06-20
Add calling application user ID to Teacher Auth analytics events

gunndabad:
We need to know what the calling service is for analytics events coming from Teacher Auth so that we can get service-specific stats. This adds an `application_user_id` entry to each web request event's `data`.


gunndabad:
[Add an identifier for the calling service to Teacher Auth analytics events](https://trello.com/c/GZ25pMHs/220-add-an-identifier-for-the-calling-service-to-teacher-auth-analytics-events)

MrKevJoy (approved):

PR #1402, opened 2024-07-01
Add a job to delete old attachments

gunndabad:
This adds a nightly job that will delete any `attachment`s from change name/DOB requests and TRN request tasks that have not been modified for 30 days. It also creates a TRS event that contains the attachment's ID, for auditing purposes. A separate PR will follow to sync the event from CRM to TRS.

We're a little light on tests here as we can't override the created/modified timestamps on CRM entities (so we can't create easily entities that appear to be 30 days old).


gunndabad:
[Job to clear out old support tasks on DQT](https://trello.com/c/hWJPuRLD/46-job-to-clear-out-old-support-tasks-on-dqt)

MrKevJoy (approved):

PR #1406, opened 2024-07-03
Redact PII from request URLs in logs

gunndabad:
We've recently added more logging of request URLs to logit.io. Some of those URLs contain sensitive information that we don't want to log.

This PR builds on the redaction setup we've previously only had on the API and extends it for all three web apps. It also adds support for redacting segments of the URL path based on the route template parameter name e.g. `trn`.

It works by looking for query parameters and route template parameters that are in either in a fixed list of known 'PII' fields or have been explicitly added to the endpoint by an `RedactParametersAttribute`.


gunndabad:
[Ensure PII is redacted from logit.io](https://trello.com/c/ktqKn02M/239-ensure-pii-is-redacted-from-logitio)

MrKevJoy (approved):

PR #1459, opened 2024-08-12
Rename person_employments, add end date logic

hortha:
### Context

Having done analysis on the last 4 months TPS data extracts and got answers from Capita about various quirks, we are now in a position to draw a line on the initial phase of the workforce data import.

Rename the `person_employments` table in TRS to `tps_employments`.
Add a `withdrawal_confirmed` field to the `tps_employments` table (so this can be passed on to BQ in case it is needed in future).
Where the end date in the extract file is in the future, use the extract date as the last known employed date.
Amend the logic used on the TPS import to include ageing the data
Set the end date to the last known employed date when we get a W indicator
Set the end date to the last known employed date when the last known employed date is greater than 5 months after the extract date
Reset the end date to NULL if the W indicator is removed in a future extract for a given TRN / LA Code / Establishment Number / Start Date (as long as the last known TPS employed date is less than 5 months after the extract date)
Reset the end date to NULL if the last known TPS employed date for a given TRN / LA Code / Establishment Number / Start Date is less than 5 months after the extract date

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Final amendments to workforce data import](https://trello.com/c/JSnwSyQD/286-final-amendments-to-workforce-data-import)

gunndabad (approved):

PR #1468, opened 2024-08-19
Add bulk person search API

gunndabad:
This adds a POST API endpoint for bulk person lookup by TRN & DOB. Up to 500 people can be requested in one go. The singular find by last name and DOB operation has been amended to return induction status, QTS and EYTS as well so that both operations have the same response structure.

In doing this I've also refactored some of the `Get Person` operation so that some its response types can be re-used. Some of the mapping logic has moved into those types too instead of being baked into specific API operations.

I've also made `PreviousNameHelper` a non-`static` class so we don't have to use `IConfiguration` directly in all the places it's called from.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # API Changelog
 
+## 20240814
+
+### `POST /v3/persons/find`
+
+New endpoint added for bulk person lookup by TRN and date of birth.
+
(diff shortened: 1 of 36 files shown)


gunndabad:
[Bulk person search API for CTR](https://trello.com/c/rPacmN6a/337-bulk-person-search-api-for-ctr)

MrKevJoy (approved):

PR #1477, opened 2024-08-28
Handle deletes and truncates in TRS tables

gunndabad:
We're about to add workforce data to the DQT reporting DB. We occasionally truncate data in those tables so we need to support that in the sync too.

Annoyingly some of the tests around this are still timing out on CI (so they're skipped) but they are working locally for me. Given that this is temporary I'm not too concerned about fixing them.


gunndabad:
[Add workforce data to DQT reporting DB](https://trello.com/c/Xac3mqG8/363-add-workforce-data-to-dqt-reporting-db)

MrKevJoy (approved):

PR #1501, opened 2024-09-05
Add support for creating alerts against a person in the TRS database to use in tests

hortha:
### Context

Our tests currently create alerts using a DQT-like schema. This needs to be amended to use the new TRS schema.

Add a `WithAlert()`  method on `CreatePersonBuilder` and use a builder type (like MQs).

All properties should be optional.

If the alert type is not specified, choose a random alert type.

If the details property is not specified, use some lorem ipsum (`Faker.Lorem.Paragraph()`)

If the external link property is not specified, generate a fake URL.

If start date is not specified, generate a random date (see MQs).

If end date is not specified, use `null`.

If end is specified but is before start date, throw an exception.

Leave the existing `WithSanction()` method for now until we have fully migrated alerts from DQT to TRS.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/ReferenceDataCache.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/ReferenceDataCache.cs
@@ -1,10 +1,15 @@
+using TeachingRecordSystem.Core.DataStore.Postgres;
+using TeachingRecordSystem.Core.DataStore.Postgres.Models;
+using TeachingRecordSystem.Core.Dqt;
 using TeachingRecordSystem.Core.Dqt.Queries;
 
-namespace TeachingRecordSystem.Core.Dqt;
+namespace TeachingRecordSystem.Core;
 
-public class ReferenceDataCache : IStartupTask
+public class ReferenceDataCache(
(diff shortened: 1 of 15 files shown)


hortha:
[Update TestData for TRS alerts schema](https://trello.com/c/sQ2odv6G/384-update-testdata-for-trs-alerts-schema)

TeachingRecordSystem/src/TeachingRecordSystem.Core/ReferenceDataCache.cs
```diff
@@ -1,10 +1,15 @@
...
+using TeachingRecordSystem.Core.Dqt;
 using TeachingRecordSystem.Core.Dqt.Queries;
 
-namespace TeachingRecordSystem.Core.Dqt;
+namespace TeachingRecordSystem.Core;
 
-public class ReferenceDataCache : IStartupTask
+public class ReferenceDataCache(
+    ICrmQueryDispatcher crmQueryDispatcher,
+    TrsDbContext dbContext) : IStartupTask
```
gunndabad, line 10: Think this’ll need to be `IDbContextFactory` as this type is used as a singleton.

PR #1506, opened 2024-09-12
Amend API responses for TRS alerts data model

gunndabad:
This updates the `vNext` API version to include alerts with the TRS data model. It also removes any references to `sanctions`. We'll need to know the 'API visibility' for each alert type (and apply some filtering) before this can be released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # API Changelog
 
+## Unreleased
+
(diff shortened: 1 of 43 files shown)


gunndabad:
[Create API version with a TRS alert data model](https://trello.com/c/WzlHPb5K/395-create-api-version-with-a-trs-alert-data-model)

MrKevJoy (approved):

PR #1533, opened 2024-09-27
Tweaks to edit start date validation

hortha:
### Context

We need to build out the front end of the TRS console to enable users to interact with Alerts data through the TRS console.
Following clarification with the business, we do not want to allow the start date to be set in the future.
Also, start date can only be amended where there is no end date (i.e. only amended for active alerts).

If there is an end date for the alert return 400 Bad Request.

Add validation to ensure start date is not in the future.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Additional validation on edit alert start date journey in TRS console](https://trello.com/c/3JyAL6Qy/437-additional-validation-on-edit-alert-start-date-journey-in-trs-console)

gunndabad (approved):

PR #1550, opened 2024-10-03
Add navigation properties to Alert models

gunndabad:
This adds navigation properties from `Alert` to `AlertType` and from `AlertType` to `AlertCategory` so we can more easily write queries that return alerts, their type and their category.
I've added an `Alerts` property to `Person` (and `Qualifications`) too.

In addition this adds `EntityFrameworkCore.Projectables` so that we can have queries that include computed properties but run on the DB. In this case we're using it for a new `IsActive` property on `Alert`.


gunndabad:
[Amend API to get alerts from TRS](https://trello.com/c/av2nYXWk/452-amend-api-to-get-alerts-from-trs)

hortha (approved):

PR #1551, opened 2024-10-03
Batch MQ and Alert creation in TestData

gunndabad:
Currently when we add an MQ or Alert in `TestData.CreatePerson()` we create each in its own transaction. This moves everything to use a single `DbContext` and it reduces the roundtrips by batching by model type.

This also re-queries the added data so that we get navigation properties loaded.


hortha (approved):

gunndabad:
[Amend API to get alerts from TRS](https://trello.com/c/av2nYXWk/452-amend-api-to-get-alerts-from-trs)

PR #1553, opened 2024-10-04
Implemented delete alert journey in TRS Console

hortha:
Implemented Delete an alert journey based on Figma designs + tests.


TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/AlertDeletedEvent.cs
```diff
@@ -0,0 +1,13 @@
+using TeachingRecordSystem.Core.Events.Models;
+using File = TeachingRecordSystem.Core.Events.Models.File;
+
+namespace TeachingRecordSystem.Core.Events;
+
+public record class AlertDeletedEvent : EventBase, IEventWithPersonId, IEventWithAlert, IEventWithKey
+{
+    public string? Key { get; init; }
```
gunndabad, line 8: Don't need a `Key` (`Key` is only used by the sync process to ensure it doesn't create multiple events for the same thing).

gunndabad (approved):

PR #1559, opened 2024-10-07
Add 'Known duplicate record' alert type

gunndabad:
This also amends the schema around `DisplayOrder` such that order is within the Alert Category and is not required unless the type is active.


gunndabad:
[Add new alert type value](https://trello.com/c/vuI6XAiL/456-add-new-alert-type-value)

MrKevJoy (approved):

PR #1572, opened 2024-10-11
Only use impersonation for DQT write actions

gunndabad:
Currently the TRS Console requires each user to have an account in DQT so that we can use impersonation in API calls. We do this so that data changes have a ‘real’ user’s name in the audit history instead of a generic ‘TRS’ user. However, relatively few pages in the console are writing data to DQT.

This change adds a `WithDqtUserImpersonation()` extension method that can be used to selectively opt-in to user impersonation per-call. It's used in the few places where we write data to DQT only.


gunndabad:
[Remove requirement for DQT account for the TRS Console](https://trello.com/c/H5rwJ6FQ/472-remove-requirement-for-dqt-account-for-the-trs-console)

MrKevJoy (approved):

PR #1598, opened 2024-10-22
Add nationalInsuranceNumber parameter to Get Person endpoint

gunndabad:
This adds an optional `nationalInsuranceNumber` parameter to the `/v3/persons/<trn>` endpoint (in `vNext`). If it's specified, it must match the record's National Insurance number (similar to the `dateOfBirth` parameter).

I couldn't decide what should happen if both a DOB and NINO are specified in a request so for now it's a `Bad Request` error. We can revisit later, if required.


gunndabad:
[Amend Get Person endpoint to support passing in a NINO](https://trello.com/c/jjAu1cJY/499-amend-get-person-endpoint-to-support-passing-in-a-nino)

MrKevJoy (approved):

PR #1657, opened 2024-11-06
Consider NINO when looking for duplicates

gunndabad:
Updates the duplicate checking process to include NINO. If NINO alone is matched it's flagged as a potential duplicate (for manual review). If NINO and DOB match it's considered a definite duplicate and we return the record's TRN and mark the request as Completed.


gunndabad:
[Update TRN allocation ID to consider NINO](https://trello.com/c/fDcGBnGf/541-update-trn-allocation-id-to-consider-nino)

MrKevJoy (approved):

PR #1660, opened 2024-11-07
Change ICurrentClientIdProvider to ICurrentUserProvider

gunndabad:
`ICurrentClientIdProvider` is a hangover from before we had any notion of application users in TRS. The replaces it with `ICurrentUserProvider`, which returns the `Guid` user ID instead of a `string`.

The `TrnRequests` table still has a `string` `ClientId` but since that'll be going away soon I've left it as-is.


gunndabad:

hortha (approved):

PR #1661, opened 2024-11-07
Link One Login user when TRN request is Completed

gunndabad:
Apply for QTS need to be able to pass us a One Login User ID with a TRN allocation request so that we can associate the One Login user with the resolved teaching record. This means that the user won't have to go through verification or matching when they sign in to AYTQ (with One Login).

This change adds a `trn_request_metadata` table in which we will stash the One Login user ID for a given request, if we have one. (This isn't going into the `trn_requests` table since we're not using that for new requests.)

Ideally we would link the One Login User to the teaching record at the point the support task is resolved but we don't have a simple and reliable way of doing that while support tasks are still managed in DQT. This puts the logic into our Get TRN request endpoint so that we can guarantee the link will have been made at the point we return the TRN to the calling service. Once support tasks are moved to TRS we can move the work to happen on support task completion.

A follow-up PR will make the change to actually write to the `trn_request_metadata` table. That same PR will cover the case where the TRN request is immediately resolved (since the calling service won't ever poll the `GET` endpoint in those cases and we still need the link to have been made).

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Api/TrnRequestHelper.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Api/TrnRequestHelper.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using TeachingRecordSystem.Core.DataStore.Postgres;
 using TeachingRecordSystem.Core.DataStore.Postgres.Models;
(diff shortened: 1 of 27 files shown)


gunndabad:

MrKevJoy (approved):

PR #1669, opened 2024-11-11
[DO NOT MERGE] Support passing a verified One Login user to a TRN request

gunndabad:
This change allows passing a verified One Login user ID (subject) in a TRN request.

The way this has to work today is a little convoluted since TRN requests live in DQT but One Login user information lives in TRS. Since we can't write to both systems at the same time reliably, we use the outbox pattern to write a message to CRM in the same transaction at the TRN request. A background process in TRS then polls for outbox messages and processes them, in this case updating either the `one_login_users` table with the resolved TRN and user ID or writing to the `trn_requests_metdata` table so that when the GET endpoint is polled and the TRN request is resolved the link between the TRN and One Login user can be made. (That change to the GET endpoint was made in a previous PR.)

Note this cannot be merged until the corresponding CRM change is in production that has the `dfeta_TrsOutboxMessage` entity.


gunndabad:

MrKevJoy (approved):

PR #1683, opened 2024-11-15
Add metadata to TRN requests

gunndabad:
Previous changes added metadata to TRN requests from the v3 API (via an outbox message) that contained the verified One Login user subject. When the request was resolved, the user's ID was inserted into the `one_login_users` table, marked as verified and linked to the created teaching record. This means that One Login users who are 'externally verified' and whose TRN is known don't need to verify with One Login or go through our matching process.

However, we need to back-port this to the v2 API (since AfQTS won't be able to use the v3 API before they go live with One Login) and we also need to support matching on emails, since many AfQTS users won't have a One Login user ID. The approach above of updating the `one_login_users` table when requests are resolved becomes considerably more complex with these additional requirements in mind (and it wasn't simple before).

This PR changes how things work. Now, for all TRN requests, whether v2 or v3, we write metadata that contains the One Login user ID (where we have one) and a flag that indicates whether the user has been pre-verified. The matching logic in Teacher Auth is amended to look at the TRN request metadata to attempt to match on both One Login user ID and email.


gunndabad:
[Add ID verified flag to TRN allocation API](https://trello.com/c/4GDKR4tH/576-add-id-verified-flag-to-trn-allocation-api)

MrKevJoy (approved):

PR #1684, opened 2024-11-17
Add the calling app's name to TRN allocation tasks

gunndabad:
Currently there's no way to tell which service a TRN request came from so these tasks are difficult to prioritise. This adds the calling app's name to the task's category.


gunndabad:
[Distinct duplicate task description so SET requests are distinct](https://trello.com/c/V3j4vJRu/579-distinct-duplicate-task-description-so-set-requests-are-distinct)

MrKevJoy (approved):

PR #1687, opened 2024-11-18
Add WebhookSender

gunndabad:
This adds a class that can send a `WebhookMessage` as a [CloudEvent](https://cloudevents.io/) using JSON. A follow up PR will add an HTTP message signature to this.

I've had to move a few types around into Core so that they can be used by this type.


gunndabad:
[Create a component to send a CloudEvent message over HTTP](https://trello.com/c/17QKBOiz/582-create-a-component-to-send-a-cloudevent-message-over-http)

hortha (approved):

PR #1689, opened 2024-11-19
Add HTTP message signature to webhook messages

gunndabad:
We're adding an [HTTP message signature](https://www.rfc-editor.org/rfc/rfc9421.html) to webhook messages so that receivers can confirm the message has come from us.

This updates `WebhookSender` to add the signature and its tests to check the signature can be verified successfully.

Most of the ceremony here is around DI since the provided extension methods in NSign don't allow scoping services to a specific `HttpClient` instance; I don't want things like `AddContentDigestOptions` to be around globally.


gunndabad:
[Add HTTP message signature to webhook messages](https://trello.com/c/E6kKgOEX/583-add-http-message-signature-to-webhook-messages)

MrKevJoy (approved):

PR #1692, opened 2024-11-20
Add a webhook-jwks endpoint

gunndabad:
Receivers of webhook messages need to verify they've come from us. To do that they need access to our public keys so they can check messages have been signed by us. This exposes a `/webhook-jwks` endpoint to allow just that. The `kid` parameter here aligns with the key ID uses in the HTTP message signature.


gunndabad:
[Add jwks endpoint for webhook signing keys](https://trello.com/c/Qg7MHDx2/589-add-jwks-endpoint-for-webhook-signing-keys)

MrKevJoy (approved):

PR #1699, opened 2024-11-21
Add a background service for delivering webhook messages

gunndabad:
This adds a `BackgroundService` that polls for due `WebhookMessages` and tries to send them with `WebhookSender`. If that's successful the message is marked as delivered. If not it's scheduled to be retried (unless we've run out of retries).

We query in relatively small batches within a transaction to try to keep the transaction relatively short. We send messages in parallel for the same reason. The `for update skip locked` means that if multiple instances of this service are running in parallel they won't be contending for the same messages.


gunndabad:
[Create webhook message sending service](https://trello.com/c/6VUZ7l44/590-create-webhook-message-sending-service)

MrKevJoy (approved):

PR #1718, opened 2024-11-29
Fix migrate-db command to work with missing database

gunndabad:
The `alter user...` statement won't work unless the DB exists but it's the `MigrateAsync` that ensures the DB exists. This reorders the statements.


CathLass (approved):
I don't know what "alter user {user} with replication" does but I can create the trs DB now with the order changed

PR #1719, opened 2024-11-29
Add methods to Person for changing induction

gunndabad:
This updates the `Person` model to add `CpdInduction*` fields. We're tracking 'induction according to CPD' separately so that we can correctly infer the right status in all cases (since CPD doesn't know about all induction statuses).


gunndabad:
[Build API to write induction data to TRS](https://trello.com/c/cEnLfZ5w/609-build-api-to-write-induction-data-to-trs)

PR #1727, opened 2024-12-04
Require start date and completed date when InductionStatus is Failed

gunndabad:
I previously thought that `Failed` shouldn't come with either a start date or completion date but it should have both.

### DIFF
--- a/TeachingRecordSystem/tests/TeachingRecordSystem.TestCommon/TestData.CreatePerson.cs
+++ b/TeachingRecordSystem/tests/TeachingRecordSystem.TestCommon/TestData.CreatePerson.cs
@@ -1082,13 +1082,13 @@ public partial class TestData
 
                 var status = _status.ValueOrFailure();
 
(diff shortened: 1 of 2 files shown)


TeachingRecordSystem/tests/TeachingRecordSystem.TestCommon/TestData.CreatePerson.cs
```diff
@@ -1124,12 +1124,12 @@ public CreatePersonInductionBuilder WithCompletedDate(DateOnly? completedDate)
                 var status = _status.ValueOrFailure();
                 var startDate = _startDate.ValueOrFailure();
 
-                if (status is not InductionStatus.Passed && completedDate is not null)
+                if (status is not InductionStatus.Passed and not InductionStatus.Failed && completedDate is not null)
                 {
                     throw new InvalidOperationException($"Completed date cannot be non-null when the status is {status}.");
```
hortha, line 1129: would that be better as "must be null" ?

hortha (approved):

PR #1779, opened 2024-12-31
Induction edit start date

CathLass:
### Context

Edit start date https://trello.com/c/wl77dRCF/629-inductions-dev-select-start-date

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Persons/PersonDetail/EditInduction/StartDate.cshtml.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Persons/PersonDetail/EditInduction/StartDate.cshtml.cs
@@ -2,6 +2,7 @@ using System.ComponentModel.DataAnnotations;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Filters;
 using TeachingRecordSystem.Core.DataStore.Postgres;
+using TeachingRecordSystem.Core.DataStore.Postgres.Models;
 
 namespace TeachingRecordSystem.SupportUi.Pages.Persons.PersonDetail.EditInduction;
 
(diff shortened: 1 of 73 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Persons/PersonDetail/EditInduction/StartDate.cshtml.cs
```diff
@@ -23,19 +35,36 @@ public string BackLink
...
     public async Task<IActionResult> OnPostAsync()
     {
+        if (StartDate > _clock.Today)
+        {
+            ModelState.AddModelError(nameof(StartDate), "The induction start date cannot be in the future");
+        }
+        if (StartDate < new DateOnly(1999, 5, 7)) // CML TODO - business logic - move
+        {
+            ModelState.AddModelError(nameof(StartDate), "The induction start date cannot be before 7th May 1999");
+        }
```
CathLass, line 58: Validation logic to be added to `person` ? Not sure I can add it to the `ValidateInductionData` method because that's being used for updating the cpd induction data which I guess might have dates prior to 7th may 1999?

gunndabad: We could put a `staticDateOnly EarliestInductionStartDate` on `Person`; we don't have a great way of sharing validation logic with Razor Pages currently.

CathLass: done - I might get in trouble for this though because that error string is supposed to say 7th May not 7 May. I propose to deal with that if it's mentioned and not before

The method ValidateInductionData - is that intended to be only for the API? I know this FE is updating the DB directly but I expect I'll call a validate method before I update the record from the last page of the wizard.

gunndabad: We have `UiDefaults.DateOnlyDisplayFormat` to use for date formatting. 

>method ValidateInductionData - is that intended to be only for the API?
Yes, we don't have a way of easily sharing validation logic across the API and the front-end currently.

gunndabad (approved):

PR #1797, opened 2025-01-08
Remove DQT sync columns for migrated entities

gunndabad:
These attributes were added just in case we needed them around migration. We didn't, so now they're going.

We still have a few around `Person` but these can be tidied up later when induction migration changes are less frequent.


TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
```diff
@@ -79,48 +79,6 @@ public async Task SyncAuditAsync(
         await Task.WhenAll(audits.Select(async kvp => await auditRepository.SetAuditDetailAsync(entityLogicalName, kvp.Key, kvp.Value)));
     }
 
-    public static Alert? MapAlertFromDqtSanction(
```
gunndabad, line 82: I missed this when removing alert-related migration code.

PR #1813, opened 2025-01-16
Induction exemption reason

CathLass:
### Context

https://trello.com/c/BIn10hhQ/665-inductions-dev-select-exemption-reason

Edit exemption reason - when induction status is exempt

The category headings for the exemption reasons code will be replaced when the categories are stored in the DB, which will be done in a later PR


gunndabad:
>Q1. The category headings for the exemption reasons was quickly bashed-together code to be replaced when the categories are stored in the DB - not sure whether it's worth refining (depending on when the DB work will be done)

We can do that in a separate PR

>Q2. The cache returns the postgres models - I'm copying its use as it's being done similarly elsewhere for now, but I wonder is there any value in the cache returning non-active reasons? The FE doesn't want all of the ExemptionReasons postgres model so I was thinking of maybe putting a more slimline model in between if the cache behaviour itself needs to return the whole thing

If we only want a subset of the exemption reasons (active only, for example) then we should add a parameter to the `Get*` method on `ReferenceDataCache` like `GetAlertTypesAsync(bool activeOnly = false)`

>Q3. I'm thinking what I've missed on all the pages so far is a safeguard that they can't be invoked if they aren't valid for the Induction state rather than relying on not being able to navigate to them. E.g. this page is invalid if the InductionStatus != Exempt. Thoughts?

Yes, the pages shouldn't be available if the journey state means they aren't valid.

hortha (approved):

PR #1814, opened 2025-01-16
Add UK to Country reference data set

gunndabad:
We don't know the full set of reference data set, but having one country unblocks building route UI.


gunndabad:
[Add country model and reference data](https://trello.com/c/mxOX34Ip/710-add-country-model-and-reference-data)

MrKevJoy (approved):

PR #1823, opened 2025-01-21
Fix Induction event model mapping from Person

gunndabad:


TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/Models/Induction.cs
```diff
@@ -21,8 +21,8 @@ public record Induction
         CompletedDate = person.InductionCompletedDate,
         ExemptionReasonIds = person.InductionExemptionReasonIds,
         CpdStatus = person.CpdInductionStatus.ToOption(),
-        CpdStartDate = Option.Some(person.CpdInductionStartDate),
-        CpdCompletedDate = Option.Some(person.CpdInductionCompletedDate),
+        CpdStartDate = person.CpdInductionStatus is not null ? Option.Some(person.CpdInductionStartDate) : default,
```
hortha, line 24: Is there a reason why they can't both just be ToOption() ?

hortha: scratch that .. I see that it's actually checking the status value

gunndabad: > Is there a reason why they can't both just be ToOption() ?

That won't compile - the types are different.

hortha (approved):

PR #1860, opened 2025-02-07
Added Job to import ITT providers

hortha:
### Context

We need a set of ITT providers that is shared with Register and we can associate with professional statuses in TRS.

Add a UKPRN field to the `training_providers` table in TRS.
Add a nightly job that calls the publish API and upserts the records into the TRS table. 
Note that some providers should be explicitly excluded (see the linked ticket).
Added trs_training_providers table to the reporting db.


TeachingRecordSystem/src/TeachingRecordSystem.Core/Jobs/HostApplicationBuilderExtensions.cs
```diff
@@ -235,6 +235,11 @@ public static IHostApplicationBuilder AddBackgroundJobs(this IHostApplicationBui
                     job => job.ExecuteAsync(/*dryRun: */true, CancellationToken.None),
                     Cron.Never);
 
+                recurringJobManager.AddOrUpdate<RefreshTrainingProvidersJob>(
+                    nameof(RefreshTrainingProvidersJob),
+                    job => job.ExecuteAsync(CancellationToken.None),
+                    Cron.Never);
```
gunndabad, line 241: Don’t we  want a schedule for this?

gunndabad (approved):

PR #1899, opened 2025-03-03
Routes save data

CathLass:
### Context

Save changes and send event


TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/ProfessionalStatusUpdatedEvent.cs
```diff
@@ -0,0 +1,35 @@
...
+    public required string? ChangeReasonDetail { get; init; }
+    public required File? EvidenceFile { get; init; }
+    public required ProfessionalStatusUpdatedEventChanges Changes { get; init; }
+}
+
+[Flags]
+public enum ProfessionalStatusUpdatedEventChanges
+{
+    None = 0,
+    ProfessionalStatus = 1 << 0, // CML TODO needed?
```
CathLass, line 22: I'm unclear about this ProfessionalStatus, which is derived from the route, but also stored in the model.

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Infrastructure/Filters/CheckProfessionalStatusExistsFilter.cs
```diff
@@ -23,9 +23,15 @@ public async Task OnResourceExecutionAsync(ResourceExecutingContext context, Res
...
 
-        var currentProfessionalStatus = await dbContext.ProfessionalStatuses
+        var request = dbContext.ProfessionalStatuses
             .FromSql($"select * from qualifications where qualification_id = {qualificationId} for update") // https://github.com/dotnet/efcore/issues/26042
             .Include(ps => ps.Person)
+            .Include(ps => ps.TrainingProvider)
+            .Include(ps => ps.TrainingCountry)
+            .Include(ps => ps.Route)
+            .Include(ps => ps.InductionExemptionReason);
+
```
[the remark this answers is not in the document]
gunndabad: Should be fine, I think

gunndabad:
>When I update the InductionExemptionReason from the route, I'm just updating the ProfessionalStatus table, not the exemption reason in person - is that being dealt with later?

Yes

> I'm unclear about what I should be doing with ProfessionalStatusType, which is derived from the route, and so exists in the RouteToProfessionalStatus part of the ProfessionalStatus postgres model), but is also stored directly in that model.

I'll make a note to remove `ProfessionalStatusType` from `ProfessionalStatus`; we can pull it from the route instead.

TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/Models/ProfessionalStatus.cs
```diff
@@ -0,0 +1,35 @@
...
+    public required RouteToProfessionalStatus Route { get; init; }
+    public required ProfessionalStatusStatus Status { get; init; }
+    public required DateOnly? AwardedDate { get; init; }
+    public required DateOnly? TrainingStartDate { get; init; }
+    public required DateOnly? TrainingEndDate { get; init; }
+    public required Guid[] TrainingSubjectIds { get; init; }
+    public required TrainingAgeSpecialismType? TrainingAgeSpecialismType { get; init; }
+    public required int? TrainingAgeSpecialismRangeFrom { get; init; }
+    public required int? TrainingAgeSpecialismRangeTo { get; init; }
+    public required Country? TrainingCountry { get; init; }
```
gunndabad, line 16:
```suggestion
    public required Guid? TrainingCountryId { get; init; }
```

TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/ProfessionalStatusUpdatedEvent.cs
gunndabad, line 22: Don't need this

gunndabad (approved):

PR #1916, opened 2025-03-20
route exemption yes-no page plus related data changes

CathLass:
### Context
Induction exemption page - does this person have an exemption.

Data changes:
- allow for the mapping of exemption reasons from routes
- store which exemptions are implicit (i.e. the " does this person have an exemption?" question does not apply)
- rename of FieldRequirement.**NotRequired** to **NotApplicable**
- some routes e.g. "High Potential ITT" seem to have changed from having an exemption to not (going from the latest spreadsheet version)


TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Models/ProfessionalStatus.cs
```diff
@@ -26,8 +26,7 @@ public ProfessionalStatus()
     public Country? TrainingCountry { get; }
     public required Guid? TrainingProviderId { get; set; }
     public TrainingProvider? TrainingProvider { get; }
-    public required Guid? InductionExemptionReasonId { get; set; }
-    public InductionExemptionReason? InductionExemptionReason { get; }
+    public required bool? ExemptFromInduction { get; set; }
```
CathLass, line 29: I made it nullable because it doesn't apply when the status isn't 'Approved' or 'Awarded', rather than it being false

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/RoutesToProfessionalStatus/EditRoute/InductionExemption.cshtml.cs
```diff
@@ -0,0 +1,76 @@
...
+    public async override Task OnPageHandlerExecutionAsync(PageHandlerExecutingContext context, PageHandlerExecutionDelegate next)
+    {
+        Route = await referenceDataCache.GetRouteToProfessionalStatusByIdAsync(JourneyInstance!.State.RouteToProfessionalStatusId);
+        if (Route.InductionExemptionReasonId is not null)
+        {
+            var exemptionReason = await referenceDataCache.GetInductionExemptionReasonByIdAsync(Route.InductionExemptionReasonId!.Value);
+            if (exemptionReason.RouteImplicitExemption)
+            {
+                //context.Result = new NotFoundResult();
+                throw new InvalidOperationException("This route provides an implict induction exemption");
```
CathLass, line 61: This is a page guard for when it doesn't apply to the route/status combo - I threw an exception to be handled by the exception page handler, but do we want just the 'Not Found' instead?

TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/Models/ProfessionalStatus.cs
```diff
@@ -13,7 +13,7 @@ public record ProfessionalStatus
     public required int? TrainingAgeSpecialismRangeTo { get; init; }
     public required string? TrainingCountryId { get; init; }
     public required Guid? TrainingProviderId { get; init; }
-    public required Guid? InductionExemptionReasonId { get; init; }
+    public bool? ExemptFromInduction { get; set; }
```
gunndabad, line 16: Probably need a migration for the reporting DB for this change too

CathLass: Right - is that something I missed when I did the degree type stuff and you rectified here https://github.com/DFE-Digital/teaching-record-system/commit/1124c3105ce26dc9392ae18b6d3d6c3a7ee3ef00

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/RoutesToProfessionalStatus/EditRoute/InductionExemption.cshtml.cs
gunndabad: BadRequest, I'd say

gunndabad (approved):

PR #1924, opened 2025-03-31
Populate dfeta_allowpiiupdatesfromregister based on AllowContactPiiUpdatesFromUserIds config setting

hortha:
### Context

There are rules in place that control whether we accept PII updates from Register over the API. Those rules rely on the value of the dfeta_allowpiiupdatesfromregisterattribute on the contact. The new TRN allocation API needs to be extended to populate that attribute.

Populate the dfeta_allowpiiupdatesfromregister attribute when creating a contact in the v3 TRN allocation endpoint. Its value should be true when the calling user ID is present in the AllowContactPiiUpdatesFromUserIds configuration array.

### DIFF
--- a/TeachingRecordSystem/tests/TeachingRecordSystem.Api.Tests/Startup.cs
+++ b/TeachingRecordSystem/tests/TeachingRecordSystem.Api.Tests/Startup.cs
@@ -2,13 +2,20 @@ namespace TeachingRecordSystem.Api.Tests;
 
 public class Startup
 {
-    public void ConfigureHost(IHostBuilder hostBuilder) =>
+    public void ConfigureHost(IHostBuilder hostBuilder)
+    {
         hostBuilder
             .ConfigureHostConfiguration(builder => builder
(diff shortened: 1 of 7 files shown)


TeachingRecordSystem/tests/TeachingRecordSystem.Api.Tests/Startup.cs
```diff
@@ -2,13 +2,20 @@ namespace TeachingRecordSystem.Api.Tests;
...
 {
-    public void ConfigureHost(IHostBuilder hostBuilder) =>
+    public void ConfigureHost(IHostBuilder hostBuilder)
+    {
         hostBuilder
             .ConfigureHostConfiguration(builder => builder
                 .AddUserSecrets<HostFixture>(optional: true)
-                .AddEnvironmentVariables())
+                .AddEnvironmentVariables()
+                .AddInMemoryCollection(
```
gunndabad, line 11: Can we not add this to `appsettings.Testing.json` instead?

gunndabad (approved):

PR #1929, opened 2025-04-03
Update user page authorization to use policies instead of roles

stusherwin:
Updates user page authorization filter and conventions to use the `UserManagement` policy instead of the `Administrator` role

**To clarify:** these were the only instances within `TeachingRecordSystem.SupportUi` where roles were used instead of policies - there were other instances within `TeachingRecordSystem.Api` but those were out of scope of this PR


gunndabad (approved):

hortha (approved):

stusherwin:
@gunndabad @hortha updated to add a new policy from the roles defined in `Conventions` - I missed this the first time around

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Infrastructure/Security/AuthorizationPolicies.cs
```diff
@@ -11,5 +11,6 @@ public static class AuthorizationPolicies
     public const string NonDbsAlertFlag = "NonDbsAlertFlag";
     public const string NonDbsAlertRead = "NonDbsAlertRead";
     public const string NonDbsAlertWrite = "NonDbsAlertWrite";
+    public const string AlertReadWrite = "AlertReadWrite";
```
gunndabad, line 14: Maybe just `AlertWrite` if it's only used for the Add Alert journey?

PR #1930, opened 2025-04-03
Amended workforce data import to trim whitespace from email addresses

hortha:
### Context

The workforce data extract has been amended to include (and validate) the estblishment email address however when it was run in production it appears that some of the data has whitespace for the email address rather than just being empty - this is currently being rejected as invalid but really it could be trimmed and treated as null.

Amend the workforce data import to trim any whitespace from the establishment_email_address before validating it (empty is allowed).
Applied the same logic for member_email_address too.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/WorkforceData/TpsCsvExtractFileImporter.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/WorkforceData/TpsCsvExtractFileImporter.cs
@@ -141,12 +141,16 @@ public class TpsCsvExtractFileImporter(
                 loadErrors = loadErrors | TpsCsvExtractItemLoadErrors.GenderIncorrectFormat;
             }
 
-            if (row.MemberEmailAddress is not null && !EmailAddress.TryParse(row.MemberEmailAddress, out _))
+            var memberEmailAddress = row.MemberEmailAddress?.Trim();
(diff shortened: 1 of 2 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/WorkforceData/TpsCsvExtractFileImporter.cs
```diff
@@ -141,12 +141,16 @@ INSERT INTO tps_csv_extracts (
                 loadErrors = loadErrors | TpsCsvExtractItemLoadErrors.GenderIncorrectFormat;
             }
 
-            if (row.MemberEmailAddress is not null && !EmailAddress.TryParse(row.MemberEmailAddress, out _))
+            var memberEmailAddress = row.MemberEmailAddress?.Trim();
+            memberEmailAddress = memberEmailAddress?.Length > 0 ? memberEmailAddress : null;
+            if (memberEmailAddress is not null && !EmailAddress.TryParse(memberEmailAddress, out _))
```
stusherwin, line 146: Would the following be more succinct?
```
if (!memberEmailAddress.IsNullOrWhiteSpace() && !EmailAddress.TryParse(memberEmailAddress, out _))
```

hortha: I hear you but I still need the trimmed value to use later though hence trimming it first and using a variable

PR #1939, opened 2025-04-10
Add Api.UnitTests project

gunndabad:


TeachingRecordSystem/tests/TeachingRecordSystem.Api.UnitTests/DummyTest.cs
```diff
@@ -0,0 +1,8 @@
+namespace TeachingRecordSystem.Api.UnitTests;
+
+public class DummyTest
+{
+    // Dummy test so that CI passes
+    [Fact]
+    public void AlwaysPasses() { }
```
gunndabad, line 7: This will be removed when we have tests added but we need at least one to make CI happy.

TeachingRecordSystem/tests/TeachingRecordSystem.Api.UnitTests/GlobalUsings.cs
```diff
@@ -0,0 +1 @@
+//global using TeachingRecordSystem.TestCommon;
```
gunndabad, line 1: Commented out to stop build warning; it'll be restored once we have some tests that use it.

MrKevJoy (approved):

PR #1953, opened 2025-04-15
Fix tests that are dependent on outbox message processing

gunndabad:
Also moves the 'PII update' flag tests into unit tests project.


TeachingRecordSystem/tests/TeachingRecordSystem.Api.UnitTests/OperationTestFixture.cs
```diff
@@ -45,15 +51,4 @@ public OperationTestFixture(
     public DbFixture DbFixture { get; }
 
     public TestData TestData { get; }
-
-    public async Task InitializeAsync()
```
gunndabad, line 49: This method isn't invoked for fixtures initialized by DI.

MrKevJoy (approved):

PR #1954, opened 2025-04-15
Amended to default training country reference as XK in DQT when left null in API

hortha:
### Context

During testing of the new routes v3 API, a bug was found where ITT records are created with null training country.

This is due to small piece of logic not being ported from the v2 API.

Any route types allowing null in the API need to default to a United Kingdom country code (XK) in the DQT ITT record as it is mandatory and will cause issues if records need to be amended manually.


TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/SetProfessionalStatus.cs
```diff
@@ -138,10 +138,10 @@ public async Task<ApiResult<SetProfessionalStatusResult>> HandleAsync(SetProfess
         Guid? countryId = null;
         if (command.TrainingCountryReference is not null)
         {
-            var (IsSuccess, Result) = await command.TrainingCountryReference.TryConvertFromTrsCountryReferenceAsync(referenceDataCache);
+            var (IsSuccess, Result) = await command.TrainingCountryReference!.TryConvertFromTrsCountryReferenceAsync(referenceDataCache);
```
gunndabad, line 141: Nit:
```suggestion
            var (isSuccess, result) = await command.TrainingCountryReference!.TryConvertFromTrsCountryReferenceAsync(referenceDataCache);
```

gunndabad (approved):

PR #1963, opened 2025-04-22
Handle 'definite duplicate' TRN requests in GET endpoint

gunndabad:
During TRN allocation, when we've found a 'definite duplicate' (i.e. exactly one record matches with NINO and DOB), we return its TRN immediately with status `Completed`. However, if the `GET` endpoint were subsequently polled, we'd return a `404` since we have no record of the request in the `trn_requests` table (legacy) and we have no CRM `contact` with matching `dfeta_trnrequestid`.

This change adds `ResolvedPersonId` to `TrnRequestMetadata` and populates it during TRN allocation if the resolved record is known. Every time a `Pending` request is queried we check if it's been completed since the last time and populate `ResolvedPersonId` if it has been.

At some point we'll need a back-fill job to populate this field for all existing requests, since we won't have `dfeta_trnrequestid` available after the contact record has been migrated to TRS.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/TrnRequestHelper.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/TrnRequestHelper.cs
@@ -34,52 +34,70 @@ public class TrnRequestHelper(
         var dbTrnRequest = await dbContext.TrnRequests.SingleOrDefaultAsync(r => r.ClientId == applicationUserId.ToString() && r.RequestId == requestId);
         var metadata = await dbContext.TrnRequestMetadata.SingleOrDefaultAsync(r => r.ApplicationUserId == applicationUserId && r.RequestId == requestId);
 
+        Guid contactId;
+
         if (metadata is null)
         {
             return null;
         }
 
-        if (dbTrnRequest is not null)
+        if (metadata.ResolvedPersonId is Guid personId)
         {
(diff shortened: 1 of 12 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/TrnRequestHelper.cs
```diff
@@ -34,52 +34,70 @@ public string GetAccessYourTeachingQualificationsLink(string trnToken) =>
...
-            var contact = await GetContactAsync(trnRequestContact.Id);
-            return new(applicationUserId, contact, metadata);
+            if (metadata.TrnToken is null && metadata.EmailAddress is not null)
+            {
+                metadata.TrnToken = await CreateTrnTokenAsync(result.Trn, metadata.EmailAddress);
+            }
+
+            metadata.ResolvedPersonId ??= result.Contact.Id;
+
+            await dbContext.SaveChangesAsync();
```
hortha, line 93: nitpicking I guess but this method is a "Get" and it seems to be updating the DB

hortha (approved):

TeachingRecordSystem/src/TeachingRecordSystem.Core/TrnRequestHelper.cs
gunndabad: It's true, but the primary function is still retrieving a `TrnRequestResult`

PR #1980, opened 2025-04-28
Don't show empty induction audit events on the timeline

gunndabad:


gunndabad:
> I added a test

Thanks!

PR #1990, opened 2025-04-30
Prevent syncing notes that contain certain text

MrKevJoy:
https://trello.com/c/S6eTf26f/1012-amend-notes-sync-job-to-exclude-notes-that-will-not-be-migrated-from-the-dqt

Don't allow syncing of notes with terms defined in the above trello ticket.

Added a few tests, but not for every term.

- Mixed casing
- Substring
- Exact match

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master
-   [x] Run DQT integration tests locally (if appropriate)

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
@@ -118,153 +118,150 @@ public class TrsDataSyncHelper(
         await Task.WhenAll(audits.Select(async kvp => await auditRepository.SetAuditDetailAsync(entityLogicalName, kvp.Key, kvp.Value)));
     }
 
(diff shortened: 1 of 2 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
```diff
@@ -117,9 +118,160 @@ public async Task SyncAuditAsync(
         await Task.WhenAll(audits.Select(async kvp => await auditRepository.SetAuditDetailAsync(entityLogicalName, kvp.Key, kvp.Value)));
     }
 
-    private DqtNoteInfo MapNoteFromDqtAnnotation(
+    private DqtNoteInfo? MapNoteFromDqtAnnotation(
         Annotation annotation)
     {
+        List<string> GetIgnoreNotesContainingTerms()
```
gunndabad, line 124: Can we put these into a static to save allocating the same list every time?

gunndabad (approved):

PR #1996, opened 2025-05-02
Recalculate professional status dates when a route is added or updated

gunndabad:
This adds a method to `Person` that recalculates the QTS/EYTS/PQTS/EYPS attributes on Person when a given `ProfessionalStatusType` is changed. This method is called when a route is added or updated.

I've updated the `ProfessionalStatus{Created|Updated}Event`s too to capture previous and new QTS/EYTS/PQTS/EYPS attributes.

I've fixed up the `ProfessionalStatusTypes` for non-QTS routes in the route reference data too.


TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Models/Person.cs
```diff
@@ -394,6 +396,56 @@ public bool InductionStatusManagedByCpd(DateOnly now)
         return (CpdInductionModifiedOn is not null && (InductionCompletedDate is null || InductionCompletedDate > sevenYearsAgo));
     }
 
+    public bool RefreshProfessionalStatusAttributes(
+        ProfessionalStatusType professionalStatusType,
+        IReadOnlyCollection<RouteToProfessionalStatus> allRoutes,
+        IEnumerable<ProfessionalStatus>? professionalStatusesHint = null)
```
gunndabad, line 402: This parameter is here for when we're adding a professional status; when this method is called `this.Qualifications` won't have the new professional status in it yet.

PR #2021, opened 2025-05-08
Add an interface for events that can possibly have a related Person

gunndabad:
We have an `IEventWithPerson` interface that's used for extracting the `PersonId` when we're saving events so we can populate the `person_id` column. This column is indexed and exists for querying events by person efficiently.

We have a few places coming up where events of the same type _might_ relate to a person but don't always. The `Guid PersonId` spec on `IEventWithPerson` doesn't work in that case.

This adds `IEventWithOptionalPerson` where `PersonId` is nullable.

I considered making `IEventWithPerson` have a `Nullable<Guid>` instead but it implies all our existing `IEventWithPerson` events can have null `PersonId` set, which isn't valid.

Another alternative would be to use `Guid.Empty` or some other well-known GUID but that seemed a little weird.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/EventBase.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/EventBase.cs
@@ -49,4 +49,22 @@ public abstract record EventBase
 
         return (EventBase)JsonSerializer.Deserialize(payload, eventType, JsonSerializerOptions)!;
     }
+
+    public bool TryGetPersonId(out Guid personId)
+    {
+        if (this is IEventWithPersonId { PersonId: var eventPersonId })
+        {
+            personId = eventPersonId;
+            return true;
+        }
+
+        if (this is IEventWithOptionalPersonId { PersonId: { } eventOptionalPersonId })
+        {
+            personId = eventOptionalPersonId;
+            return true;
+        }
+
+        personId = Guid.Empty;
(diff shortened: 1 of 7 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/EventBase.cs
```diff
@@ -49,4 +49,22 @@ public static EventBase Deserialize(string payload, string eventName)
...
+            return true;
+        }
+
+        if (this is IEventWithOptionalPersonId { PersonId: { } eventOptionalPersonId })
+        {
+            personId = eventOptionalPersonId;
+            return true;
+        }
+
+        personId = Guid.Empty;
```
stusherwin, line 67: Shouldn't this return a nullable Guid?

hortha (approved):

TeachingRecordSystem/src/TeachingRecordSystem.Core/Events/EventBase.cs
stusherwin: The caller could then convert that to an empty Guid if needed

gunndabad: It's more jarring to use if it's nullable; this is similar to `int.TryParse()` which don't output a nullable, even when parsing fails.

PR #2023, opened 2025-05-08
Proactively sync matched Person before comparing attributes

gunndabad:
The process to resolve a TRN request uses Person data from the TRS DB. This is synced in _near_ real time but can lag. Showing stale data on these pages, particularly when the user is choosing what person data to use, is not ideal.

This adds a proactive sync to GET requests for the pages that compare the matched record to the request and show the resolved person attributes. If the sync fails for whatever reason we won't block progress; we don't want a transient error or CRM slow-ness to prevent the journey being completed when 99% of time the data won't have changed.

Once this is live we can monitor and adjust when/where we do this sync and whether the 5 second threshold is sufficient.

We don't need a sync on the PersonDetail page (that we link to once the request is resolved) since that's getting Person data from CRM so will never be showing stale data.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/PageModelExtensions.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/PageModelExtensions.cs
@@ -1,8 +1,45 @@
+using System.Transactions;
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using TeachingRecordSystem.Core.Services.TrsDataSync;
 
 namespace TeachingRecordSystem.SupportUi;
 
 public static class PageModelExtensions
 {
+    private static readonly TimeSpan _defaultSyncThreshold = TimeSpan.FromSeconds(5);
+
     public static PageResult PageWithErrors(this PageModel pageModel) => new PageResult() { StatusCode = StatusCodes.Status400BadRequest };
+
+    public static async Task SyncPersonAsync(
+        this PageModel pageModel,
+        Guid personId,
+        CancellationToken cancellationToken = default)
+    {
+        var syncHelper = pageModel.HttpContext.RequestServices.GetRequiredService<TrsDataSyncHelper>();
+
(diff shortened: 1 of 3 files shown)


MrKevJoy (approved):

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/PageModelExtensions.cs
```diff
@@ -1,8 +1,45 @@
...
+
+    public static async Task SyncPersonAsync(
+        this PageModel pageModel,
+        Guid personId,
+        CancellationToken cancellationToken = default)
+    {
+        var syncHelper = pageModel.HttpContext.RequestServices.GetRequiredService<TrsDataSyncHelper>();
+
+        using var txn = new TransactionScope(TransactionScopeOption.Suppress, TransactionScopeAsyncFlowOption.Enabled);
+        await syncHelper.SyncPersonAsync(personId, syncAudit: false, cancellationToken: cancellationToken);
```
stusherwin, line 21: Just curious, is the transaction aborted if the token is cancelled?

stusherwin: Also `SyncPersonAsyc()` creates its own transaction, assuming that's not an issue?

stusherwin: https://github.com/DFE-Digital/teaching-record-system/blob/fe62255374dca7f7e20fb15cb88d8a4be4a313a3/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs#L487

Docs for `BeginTransactionAsync()` say nested transactions are not supported

gunndabad: The `TransactionScopeOption.Suppress` is there to avoid the nested transactions problem; it's fine that it doesn't share a parent transaction.

> Just curious, is the transaction aborted if the token is cancelled?

The transaction created inside`TrsDataSyncHelper` would be, yes.

PR #2029, opened 2025-05-12
Add pagination to list of API TRN requests

gunndabad:
I've tweaked the pagination bits added for the users list:
- The `Pagination` class is gone in favour of a `GetPageAsync()` extension method on `IQueryable` and `ResultPage<T>`. With this pagination can be chained onto an existing query.
- The `PaginationViewModel` now takes in a delegate that returns the link so we can keep link generation in `TrsLinkGenerator`.

It would be good to have a reusable way to test pages that use these pagination bits but I've left that for a future PR.


TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/SupportTasks/ApiTrnRequests/Index.cshtml.cs
```diff
@@ -20,12 +24,18 @@ public class Index(TrsDbContext dbContext) : PageModel
...
+
+    public ResultPage<Result>? Results { get; set; }
+
+    public PaginationViewModel? Pagination { get; set; }
 
     public async Task<IActionResult> OnGetAsync()
     {
-        var sortDirection = SortDirection ?? SupportUi.SortDirection.Ascending;
-        var sortBy = SortBy ?? ApiTrnRequestsSortByOption.Name;
+        var sortDirection = SortDirection ??= SupportUi.SortDirection.Ascending;
```
gunndabad, line 37: I broke these two properties previously; we rely on SortDirection being not `null` to have the tag helpers do the right thing.

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Users/Index.cshtml.cs
```diff
@@ -52,22 +46,22 @@ public async Task OnGetAsync()
...
         ]);
 
-        var baseQuery = _dbContext.Users
+        var baseQuery = dbContext.Users
             .Where(u => u.UserType == UserType.Person && !string.IsNullOrWhiteSpace(u.Email))
             .Where(u => showAdminRole || u.Role != UserRoles.Administrator);
 
         var filteredQuery = filters.Apply(baseQuery);
 
-        var totalUserCount = await filters.CountAsync(filteredQuery);
```
stusherwin, line 61: Not calling `filters.CountAsync()` will break the filter counts - surprised the tests pass

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/QueryableExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        var total = await source.CountAsync();
+
+        var lastPage = ResultPage.GetLastPage(total, itemsPerPage);
+
+        // Is the page requested out of bounds?
+        if (page > lastPage)
+        {
+            page = lastPage;
+        }
```
stusherwin, line 39: This logic felt cleaner to me when it was all part of a single `Pagination` class, now it's split across many classes

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/QueryableExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public static async Task<ResultPage<T>> GetPageAsync<T>(
+        this IQueryable<T> source,
+        int? currentPage,
+        int itemsPerPage)
+    {
+        var page = Math.Max(currentPage ?? 1, 1);
+
+        var total = await source.CountAsync();
```
stusherwin, line 31: There was a reason this count was part of the `Filters` class, it calculates the total item count as part of calculating the counts for all the filter options

stusherwin (changes requested):

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Users/Index.cshtml.cs
gunndabad: Ah I see - I didn't realise `CountAsync()` had a side-effect

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/QueryableExtensions.cs
gunndabad: >now it's split across many classes

How so?

stusherwin: The `ResultsPage<T>` classes and the extension method class

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Users/Index.cshtml.cs
stusherwin: `CountAsync()` is one of the main functions of the `Filters` class - I guess it could have a better name - maybe `CalculateFilterCounts` with an out parameter int `totalItemCount` - that would give more indication that the total count is more a by-product of a more complex operation

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/QueryableExtensions.cs
stusherwin: Also if this is used with filters, then calling `CountAsync` here will create a third query - one to calculate the filter counts, one to get the current page of items, and this one to calculate the total count

gunndabad: Yeh - I'm starting to think that the filters implementation and the simpler one here can't use the same implementation...

stusherwin: I wonder if something like a querybuilder would work where it knows whether or it has filters, and either delegates to the filter count or does a `CountAsync`

stusherwin (approved):

PR #2046, opened 2025-05-22
Prevent 's in generated test names

gunndabad:
' characters in test names cause problems in a few places e.g. O'Reilly. While we would ideally be handling these by escaping the name, for now prevent those names from being generated at all.


stusherwin:
@gunndabad doesn't this cause problems with existing names? Or are there no existing names with ' characters?

stusherwin (changes requested):
Do we not want to fix the underlying issue?

stusherwin (approved):

PR #2072, opened 2025-06-02
Add skeleton of next version of GetPerson API endpoint with routes

gunndabad:
This also removes the `NpqQualifications` and `InitialTeacherTraining` `Includes`.

I've also added some additional information to `Induction` as CTR/AYTQ will need it.

https://trello.com/c/k8jCATfI/1177-routes-v3-get-api-endpoint


TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/VNext/Controllers/PersonsController.EwcWales.cs
```diff
@@ -8,7 +8,7 @@
 namespace TeachingRecordSystem.Api.V3.VNext.Controllers;
 
 [Route("persons")]
-public class PersonsController : ControllerBase
+public partial class PersonsController : ControllerBase
```
gunndabad, line 11: This is moved into a partial class because it's likely to be stuck in `VNext` for a while and it makes promoting other endpoints simpler if this is in its own file.

MrKevJoy (approved):

PR #2185, opened 2025-07-03
Added code to move induction exemptions from person to routes during migration where appropriate

hortha:
### Context

TRS supports route-level induction exemptions whereas DQT does not. As part of migration, we need to move person-level induction exemptions onto the route where possible.

When mapping a route, if the route’s RouteToProfessionalStatusType has a non-null InductionExemptionReasonId and the Person has that same InductionExemptionReasonId in their InductionExemptionReasonIds set ExemptFromInduction to true on the route and remove the reason from the Person’s InductionExemptionReasonIds. Generate a PersonInductionUpdatedEvent too with Changes == InductionExemptionReasons.

If the route’s RouteToProfessionalStatusType has a non-null InductionExemptionReasonId and the Person does NOT have that same InductionExemptionReasonId in their InductionExemptionReasonIds set ExemptFromInduction to false (see N.B. below).

If the route’s Status is Holds and the HoldsFrom date is before 7th May 2000 and the ProfessionalStatusType of the route’s RouteToProfessionalStatusType is QualifiedTeacherStatus and the Person’s InductionExemptionReasonIds contains QualifiedBefore7May2000Id, set ExemptFromInductionDueToQtsDate to true on the mapped route and remove the reason from the Person’s InductionExemptionReasonIds. Generate a PersonInductionUpdatedEvent too with Changes == InductionExemptionReasons.

N.B. it may be possible that a single person-level induction exemption maps to a route-level exemption for multiple routes.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
@@ -17,6 +17,7 @@ using TeachingRecordSystem.Core.DataStore.Postgres;
 using TeachingRecordSystem.Core.DataStore.Postgres.Models;
 using TeachingRecordSystem.Core.Dqt;
 using TeachingRecordSystem.Core.Services.Files;
+using SystemUser = TeachingRecordSystem.Core.DataStore.Postgres.Models.SystemUser;
 
 namespace TeachingRecordSystem.Core.Services.TrsDataSync;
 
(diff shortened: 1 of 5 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
```diff
@@ -1308,92 +1309,178 @@ public async Task<int> SyncEventsAsync(IReadOnlyCollection<dfeta_TRSEvent> event
         return events.Count;
     }
 
-    private async Task<int> SyncRoutesAsync(
-        IReadOnlyCollection<RouteToProfessionalStatus> routes,
+    private async Task<Guid[]> EnsureSyncPersonsAsync(Guid[] personIds, bool ignoreInvalid = false, bool dryRun = false, CancellationToken cancellationToken = default)
+    {
+        using var dbContext = TrsDbContext.Create(trsDbDataSource);
+        var alreadySyncedPersonIds = await dbContext.Persons
```
gunndabad, line 1315: Does this need `IgnoreQueryFilters()`?

gunndabad (approved):

TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
hortha: yes otherwise it doesn't sync any inactive contacts which aren't currently in TRS

TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/TrsDataSync/TrsDataSyncHelper.cs
```diff
@@ -1308,92 +1309,178 @@ public async Task<int> SyncEventsAsync(IReadOnlyCollection<dfeta_TRSEvent> event
...
         }
 
-        var toSync = routes.ToList();
-        var modelTypeSyncInfo = GetModelTypeSyncInfo<RouteToProfessionalStatus>(ModelTypes.Route);
-        await using var connection = await trsDbDataSource.OpenConnectionAsync(cancellationToken);
-
-        do
+        // Ensure we sync any Persons that are not already synced
+        var personIds = mappings.Where(m => m.MappedResults.Any(r => r.Success)).Select(m => m.ContactId).Distinct().ToArray();
+        var syncedPersonIds = await EnsureSyncPersonsAsync(personIds, ignoreInvalid, dryRun, cancellationToken);
```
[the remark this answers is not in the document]
hortha: actually yes good point

PR #2215, opened 2025-07-11
Don't recalculate more than we need to when routes change

gunndabad:
When routes are added/updated/removed, it potentially affects the person's induction status, QTLS status and the various professional status fields (e.g. `QtsDate`). Previously we were recalculating everything whenever any route was changed. This change amends that to only recalculate what we need to.

This is more about ensuring we don't clobber potentially invalid data at unexpected times e.g. if a person's `QtsDate` is 'wrong' according to their routes then amending an EYTS route shouldn't be affecting that.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Models/RouteToProfessionalStatus.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Models/RouteToProfessionalStatus.cs
@@ -92,15 +92,22 @@ public class RouteToProfessionalStatus : Qualification
             TrainingProviderId = trainingProviderId,
             TrainingSubjectIds = trainingSubjectIds ?? []
         };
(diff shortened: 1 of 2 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Models/RouteToProfessionalStatus.cs
```diff
@@ -92,15 +92,27 @@ public static RouteToProfessionalStatus Create(
             TrainingProviderId = trainingProviderId,
             TrainingSubjectIds = trainingSubjectIds ?? []
         };
-        route.RefreshExemptFromInductionDueToQtsDate(routeType.ProfessionalStatusType);
+
+        if (routeType.ProfessionalStatusType == ProfessionalStatusType.QualifiedTeacherStatus)
+        {
+            route.RefreshExemptFromInductionDueToQtsDate();
```
hortha, line 98: this seems to be doing the same thing as line 109 unless I'm not seeing something?

PR #2240, opened 2025-07-18
Update person search to allow searching for deactivated records

gunndabad:
I've not used the same filter bits from User search as there as some behaviour differences that we want here.

There are few changes around navigation in general as well to align with the designs.

This also rewires the search to use TRS instead of DQT.

https://trello.com/c/Ig4WQ1dh/1375-update-search-to-find-and-display-deactivated-records


TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Persons/Layout.cshtml
```diff
@@ -6,27 +6,11 @@
 
 @{
     Layout = "_Layout";
-    var selectedTab = (PersonsSubNavigationTab)ViewBag.SelectedTab;
+    //var selectedTab = (PersonsSubNavigationTab)ViewBag.SelectedTab;
```
hortha, line 9: commented out or should be removed?

hortha (approved):

TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/Persons/Layout.cshtml
gunndabad: ah yeh, should be taken out

PR #2302, opened 2025-07-31
Amended DOB change to create support task in TRS (behind feature flag)

hortha:
### Context

The TRS Console currently allows users to manage Change name and change date of birth requests which are stored as “Cases” in DQT.
As part of migrating from DQT, we need to move these to be stored as support tasks in TRS.

Create a new record `ChangeDateOfBirthRequestData` in the `TeachingRecordSystem.Core.Models.SupportTaskData` namespace based on existing fields in `CreateDateOfBirthChangeRequestCommand` replacing `EvidenceFileUrl` with `EvidenceFileId`.

 Amend `CreateDateOfBirthChangeRequestHandler` to create a `SupportTask` of type `ChangeDateOfBirthRequest` populating the `Data` field with a `ChangeDateOfBirthRequestData` record populated from the incoming request (when the feature flag is enabled otherwise call the existing code which creates cases in DQT).

Considerations:
- The incoming request populates the `EvidenceFileUrl` with an expiring link to download uploaded evidence.
We need to use this URL to download the evidence and then re-upload it to the `uploads` container in Azure Blob Storage for TRS, storing the file ID returned from blob storage in the `ChangeDateOfBirthRequestData.EvidenceFileId` field.
- We need to replicate the call to send an email from the CRM plugin `CaseCreatedSendEmailPlugin` (using `INotificationSender` to send emails via Notify in TRS).
- This should be kept behind a feature flag.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/CreateDateOfBirthChangeRequest.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/CreateDateOfBirthChangeRequest.cs
@@ -1,6 +1,12 @@
+using System.Transactions;
 using Microsoft.AspNetCore.StaticFiles;
+using TeachingRecordSystem.Core.DataStore.Postgres;
+using TeachingRecordSystem.Core.DataStore.Postgres.Models;
 using TeachingRecordSystem.Core.Dqt;
-using TeachingRecordSystem.Core.Dqt.Queries;
+using TeachingRecordSystem.Core.Jobs;
+using TeachingRecordSystem.Core.Jobs.Scheduling;
+using TeachingRecordSystem.Core.Models.SupportTaskData;
+using TeachingRecordSystem.Core.Services.Files;
 
 namespace TeachingRecordSystem.Api.V3.Implementation.Operations;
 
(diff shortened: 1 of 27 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/CreateDateOfBirthChangeRequest.cs
```diff
@@ -44,17 +63,60 @@ public async Task<ApiResult<CreateDateOfBirthChangeRequestResult>> HandleAsync(C
             evidenceFileMimeType = "application/octet-stream";
         }
 
-        var (_, ticketNumber) = await crmQueryDispatcher.ExecuteQueryAsync(new CreateDateOfBirthChangeIncidentQuery()
+        var stream = await evidenceFileResponse.Content.ReadAsStreamAsync();
```
gunndabad, line 66:
```suggestion
        using var stream = await evidenceFileResponse.Content.ReadAsStreamAsync();
```

TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/CreateDateOfBirthChangeRequest.cs
```diff
@@ -44,17 +63,60 @@ public async Task<ApiResult<CreateDateOfBirthChangeRequestResult>> HandleAsync(C
...
+        var evidenceFileId = await fileService.UploadFileAsync(stream, evidenceFileMimeType);
+
+        var changeRequestData = new ChangeDateOfBirthRequestData()
         {
-            ContactId = contact.Id,
+            Trn = command.Trn,
```
gunndabad, line 71: Can we use `PersonId` instead of `Trn`?

gunndabad: Actually we probably don't need that at all since the task has the PersonId on

TeachingRecordSystem/src/TeachingRecordSystem.Core/Services/Notify/HostApplicationBuilderExtensions.cs
```diff
@@ -22,6 +23,8 @@ public static IHostApplicationBuilder AddEmail(this IHostApplicationBuilder buil
             builder.Services.AddSingleton<INotificationSender, NoopNotificationSender>();
         }
 
+        builder.Services.AddTransient<SendEmailJob>();
```
gunndabad, line 26: Do we definitely need this? We have emails going out already without this registration.

gunndabad (approved):

PR #2363, opened 2025-08-14
Add link from Person to TrnRequestMetadata

gunndabad:
This is primarily so we can track where a Person record was created from. A future PR will back-fill this from the DQT data.


stusherwin (approved):
Do we need to add anything to DQT reporting migrations?

gunndabad:
> Do we need to add anything to DQT reporting migrations?

Ah yes. One day I’ll remember that.

PR #2398, opened 2025-08-21
Fixed so that evidence shows embedded in the change request page

hortha:
In the Change Name / Date of Birth page, the uploaded evidence should render embedded on the page (as well as being able to click on a link to open in a separate tab.
This PR fixes the issue


TeachingRecordSystem/src/TeachingRecordSystem.SupportUi/Pages/ChangeRequests/EditChangeRequest/Index.cshtml.cs
```diff
@@ -31,6 +32,16 @@ public void OnGet()
...
     }
 
+    public async Task<IActionResult> OnGetEvidenceAsync()
+    {
+        using var stream = await fileService.OpenReadStreamAsync(Evidence!.FileId);
+        using var ms = new MemoryStream();
+        await stream.CopyToAsync(ms);
+        var bytes = ms.ToArray();
+
+        return File(bytes, Evidence.MimeType);
```
gunndabad, line 42: Isn’t there an overload of this that takes a Stream (to avoid pulling the whole thing into memory at once)?

hortha: ah so there is!

gunndabad (approved):

PR #2422, opened 2025-08-27
Ported TRN generation from external API to internal TRS service

hortha:
### Context

Now that DQT is (almost) gone, we can move TRN generation into TRS itself.

Copy the TRN generation logic from the TRN Generation API into TRS itself.
Create an implementation of ITrnGenerator that uses TRS’ database (be mindful of DB transactions and ensuring the allocation occurs in the same transaction as other DB changes).

NOTE: the change to register the `DbTrnGenerator` implementation of `ITrnGenerator` and ensure any calls are part of a transaction will be in a separate PR


TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Mappings/IntReturnMapping.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using TeachingRecordSystem.Core.DataStore.Postgres.Models;
+
+namespace TeachingRecordSystem.Core.DataStore.Postgres.Mappings;
+
+public class IntReturnMapping : IEntityTypeConfiguration<IntReturn>
```
gunndabad, line 6: Do we still need this in the current version of EF?

hortha: done

gunndabad (approved):

PR #2423, opened 2025-08-27
Fix find by DOB and last name query perf

gunndabad:
We've got some timeouts in Sentry from this query; use `person_search_attributes` instead.

### DIFF
--- a/TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/FindPersonByLastNameAndDateOfBirth.cs
+++ b/TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/FindPersonByLastNameAndDateOfBirth.cs
@@ -14,16 +14,15 @@ public class FindPersonByLastNameAndDateOfBirthHandler(
     {
         var matchedPersons = await DbContext.Database.SqlQueryRaw<Result>(
                 """
-                SELECT DISTINCT p.person_id FROM persons p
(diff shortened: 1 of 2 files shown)


TeachingRecordSystem/src/TeachingRecordSystem.Api/V3/Implementation/Operations/FindPersonByLastNameAndDateOfBirth.cs
```diff
@@ -14,16 +14,15 @@ public async Task<ApiResult<FindPersonsResult>> HandleAsync(FindPersonByLastName
...
-                LEFT JOIN previous_names pn ON p.person_id = pn.person_id
-                WHERE p.date_of_birth = :date_of_birth
-                AND (p.last_name = :last_name OR pn.last_name = :last_name)
-                AND p.status = 0
-                AND p.trn IS NOT NULL
+                SELECT person_id FROM person_search_attributes
+                WHERE (attribute_type = 'LastName' AND attribute_value = (:last_name COLLATE "case_insensitive")) OR
+                      (attribute_type = 'DateOfBirth' AND attribute_value = (:date_of_birth COLLATE "case_insensitive"))
+                GROUP BY person_id
+                HAVING COUNT(DISTINCT attribute_type) = 2
```
stusherwin, line 21: Do we not need `AND p.status = 0 AND p.trn IS NOT NULL` - or are those implied by being in `person_search_attributes`?

PR #2541, opened 2025-09-16
Create pen test environment

hortha:
### Context

We need a new environment for the IT Health Check that’s on the Platform test cluster so that testing can complete without disrupting other teams and services.

Create a new environment - `pentest` - on the platform test cluster.


Makefile
```diff
@@ -36,6 +36,10 @@ pre-production: test-cluster
 tps-sandbox: production-cluster
 	$(eval include global_config/tps-sandbox.sh)
 
+.PHONY: pentest
+pentest: production-cluster
```
RMcVelia, line 40: this will need to be
 pentest: platform-test-cluster
and that means you need to add the below to the Makefile
 platform-test-cluster:
       $(eval CLUSTER_RESOURCE_GROUP_NAME=s189t01-tsc-pt-rg)
       $(eval CLUSTER_NAME=s189t01-tsc-platform-test-aks)

terraform/aks/config/pentest.tfvars.json
```diff
@@ -0,0 +1,13 @@
+{
+  "environment_name": "pentest",
+  "cluster": "platform-test",
+  "namespace": "trs-pentest",
```
RMcVelia, line 4: "namespace": "development"

PR #2599, opened 2025-09-30
Add new event models

gunndabad:
Our current way of modelling events has some drawbacks; this is the start of a revised data model to address those problems.

## How it works today
We have a rule that all data changes in TRS should create an event. This is primarily so that we have an audit trail for those data changes. We decided relatively early on that we should create *exactly one* event per 'action'. There are some up-sides to this; the Change History UI has a one-to-one mapping from an event to an entry and these events are highly-specific.

## What's wrong with the current approach?
Because we can only emit one event per action, events end up being highly-specific. This is nice for the Change History UI but is problematic when trying to query for data changes (since there are multiple events for a particular piece of data changing). It also prevents us from loosely-coupling our code (since the code that creates the event needs to know about all data changes right there and then).

## Proposed new model
The proposed new model has two things; 'actions' and 'events' (names t.b.c.). Actions are highly-specific, have a type and each data change should create exactly one e.g. `ApiTrnRequestSupportTaskResolved`. An action must have one or more 'events' and these should be generic e.g. `SupportTaskUpdated`.

The Change History view would be changed to show actions, with the data coming from each action's events.

Code can then be loosely coupled using events (with an approach as shown in https://github.com/DFE-Digital/teaching-record-system/pull/2550).

This should give us the best of both worlds; highly-specific actions for where the context of why a change was made is important and events for more generic information about data changes.

## Example - `RouteToProfessionalStatusUpdatedEvent`
The `RouteToProfessionalStatusUpdatedEvent` event is replaced with a `RouteToProfessionalStatusUpdated` action with the following events:
- `RouteToProfessionalStatusUpdated` (this would have only the data from the route itself, unlike today);
- `PersonQtsUpdated` (optional, if QTS were updated);
- `PersonInductionStatusUpdated` (optional, if the induction status were updated as a result of QTS change).

## Example - `PersonsMergedEvent`
The `PersonsMergedEvent` event is replaced with a `PersonsMerged` action with the following events:
- `PersonStatusUpdated` (for the person that's deactivated);
- `PersonDetailsUpdated` (for the record that's retained, if any of its details were changed).

## Migration approach
1. Deprecate the current set of events (e.g. move to a `Legacy` namespace).
2. Replace the current event emission with actions (one by one).
3. Amend the Change History to support querying and showing actions.
4. Write jobs/DB migrations to convert the old events to new actions & events.
5. Remove the old event types & table.


gunndabad:
I'm not convinced about the 'action' word; perhaps 'operation' or 'activity' would be better.

PR #2619, opened 2025-10-08
Rename Action to Process

gunndabad:


TeachingRecordSystem/src/TeachingRecordSystem.Core/DataStore/Postgres/Models/Process.cs
```diff
@@ -0,0 +1,13 @@
+namespace TeachingRecordSystem.Core.DataStore.Postgres.Models;
+
+public class Process
+{
+    public static Guid DqtProcessUserId { get; } = Guid.Empty;
```
stusherwin, line 5: What's this field for? Do we still need it?

gunndabad: Yeh we need it - it's for processes that happened in DQT (so we can't attach a TRS user ID to them).

TeachingRecordSystem/src/TeachingRecordSystem.Core/Models/ProcessType.cs
```diff
@@ -1,6 +1,6 @@
 namespace TeachingRecordSystem.Core.Models;
 
-public enum ActionType
+public enum ProcessType
 {
     RouteToProfessionalStatusCreated = 1,
```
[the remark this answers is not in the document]
gunndabad: Yes; I'll do that as part of next PR when these actually get used. The current entries were a demo for the initial PR.

PR #3266, opened 2026-04-09
Amended Id verification support task verify step to navigate to no matches if TRN has not been entered by the citizen user

hortha:
### Context

https://github.com/DFE-Digital/teaching-record-team-project-board/issues/60


gunndabad (approved):
Do we have e2e tests covering this?

PR #3301, opened 2026-04-20
Normalize TRN and NINO in OL support task pages

gunndabad:
We allow users to enter TRNs in formats like '01/23456' and NINOs like 'ab 12 34 56 c'. This change amends the support task journeys where we play back these fields to display values in their normalized forms e.g. '0123456' and 'AB123456C', respectively.


gunndabad:
@copilot add tests for the changes in this PR

hortha (approved):

PR #3545, opened 2026-07-02
2849 - fix fix-airbyte-replication wf service_name and schedule

hopep-mobilise:
### Context

We need to add a workflow that is used to monitor and fix Postgres replication slots.
https://trello.com/c/TyRmQvgw/2849-trs-airbyte-changes

- New APP_NAME_SUFFIX env variable created as trs use a suffix on the container deployment name.
- The deployment prefix is not the service_name but the service_short.
- The schedule has been uncommented so that the workflow will run every hour between 00:00 and 09:00

### Guidance to review

The workflow has been tested manually against this branch for dev, pre-production and production with the with check slot = true so that no actions are taken.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master
-   [x] Tested by running workflow against the branch in check mode

### DIFF
--- a/.github/workflows/fix-airbyte-replication.yml
+++ b/.github/workflows/fix-airbyte-replication.yml
@@ -20,15 +20,16 @@ on:
                 options:
                     - "true"
                     - "false"
-    # schedule: # hourly every Sunday midnight until 9am
-    #     - cron: "0 0-9 * * 0"
+    schedule: # hourly every Sunday midnight until 9am
+        - cron: "0 0-9 * * 0"
 
 permissions:
     id-token: write
 
 env:
-    SERVICE_NAME: teaching-record-system
+    SERVICE_NAME: trs
     SERVICE_SHORT: trs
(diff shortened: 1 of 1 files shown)


.github/workflows/fix-airbyte-replication.yml
```diff
@@ -20,15 +20,16 @@ on:
                 options:
                     - "true"
                     - "false"
-    # schedule: # hourly every Sunday midnight until 9am
-    #     - cron: "0 0-9 * * 0"
+    schedule: # hourly every Sunday midnight until 9am
+        - cron: "0 0-9 * * 0"
```
rgwenter-mobi, line 24: Why every hour on Sunday?

rgwenter-mobi: agreed this is what's already in place so we'll go with it for now

rgwenter-mobi (approved):

PR #3564, opened 2026-07-09
Added job to delete old ID proofs + handle gracefully in UI

hortha:
### Context

https://github.com/DFE-Digital/teaching-record-team-project-board/issues/30

### DIFF
--- a/src/TeachingRecordSystem.Core/Services/Files/BlobStorageFileService.cs
+++ b/src/TeachingRecordSystem.Core/Services/Files/BlobStorageFileService.cs
@@ -51,6 +51,17 @@ public class BlobStorageFileService : IFileService
         return blobClient.GenerateSasUri(sasBuilder).ToString();
     }
 
+    public async Task<string?> TryGetFileUrlAsync(Guid fileId, TimeSpan expiresAfter)
+    {
+        var blobClient = await GetBlobClientAsync(fileId);
+        if (!await blobClient.ExistsAsync())
+        {
+            return null;
+        }
+
+        return await GetFileUrlAsync(fileId, expiresAfter);
+    }
+
     public async Task<Stream> OpenReadStreamAsync(Guid fileId)
     {
         var blobClient = await GetBlobClientAsync(fileId);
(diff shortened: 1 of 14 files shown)


src/TeachingRecordSystem.Core/Services/Files/BlobStorageFileService.cs
```diff
@@ -51,6 +51,22 @@ public async Task<string> GetFileUrlAsync(Guid fileId, TimeSpan expiresAfter)
...
     }
 
+    public async Task<string?> TryGetFileUrlAsync(Guid fileId, TimeSpan expiresAfter)
+    {
+        if (!await FileExistsAsync(fileId))
+        {
+            return null;
+        }
+
+        return await GetFileUrlAsync(fileId, expiresAfter);
```
gunndabad: Can we do this in one API instead of two?

hortha: done

src/TeachingRecordSystem.Core/Services/Files/BlobStorageFileService.cs
```diff
@@ -51,6 +51,17 @@ public async Task<string> GetFileUrlAsync(Guid fileId, TimeSpan expiresAfter)
...
 
+    public async Task<string?> TryGetFileUrlAsync(Guid fileId, TimeSpan expiresAfter)
+    {
+        var blobClient = await GetBlobClientAsync(fileId);
+        if (!await blobClient.ExistsAsync())
+        {
+            return null;
+        }
+
+        return await GetFileUrlAsync(fileId, expiresAfter);
```
gunndabad, line 62: Can we do the exists check and GET in one call to Azure?

gunndabad (approved):

src/TeachingRecordSystem.Core/Services/Files/BlobStorageFileService.cs
hortha: GenerateSasUri does not call Azure so there is only one call already .. admittedly by it calling the GetFileUrlAsync it gets the blob client a second time if you think that would be an overhead?

gunndabad: Yeh - I'd like the `blobClient.ExistsAsync()` call to go away ideally

hortha: so no calls to Azure at all?
This method is just used by the UI component to safely know that the file exists so it doesn't feel excessive to me

PR #3641, opened 2026-07-23
Migrate Edit MQ Status journey to GOV.UK Questions

gunndabad:
Follows #3637, #3638, #3639 and #3640. Fifth of the MQ journeys — only Add MQ is left after this.

## What

Migrates the **Edit MQ Status** journey from `FormFlow` to `GovUk.Questions.AspNetCore`.

## Changes

- Add `EditMqStatusJourneyCoordinator` (`[JourneyCoordinator(JourneyNames.EditMqStatus, ["qualificationId"])]`); drop FormFlow's `IRegisterJourney` from `EditMqStatusState`.
- Rewrite Index, Reason and CheckAnswers to inject the coordinator and use `AdvanceTo` / `GetBackLink` / `DeleteInstance`, with `_jid` keys and a form-field `Cancel` submit in place of the `/cancel` handler URLs.
- Rewrite `EditMqStatusLinkGenerator` onto the shared `GetJourneyPage` extension; update both person qualifications page entry links (status and end date both point at this journey).
- Seed the starting state from the qualification in the coordinator's `GetStartingState`, so `Initialized` / `EnsureInitialized` are deleted.
- **Validation → FluentValidation:**
  - Index's `if (EndDate is null) … else if (EndDate <= StartDate) …` (only when the status is Passed) becomes a `Cascade(CascadeMode.Stop)` chain gated on `.When(m => m.Status == Passed)`. The second condition is written as `!(endDate <= m.StartDate)` — the negation of the original — so a null `StartDate` still passes exactly as before. The rule is placed before the `Status` rule to keep the error-summary ordering unchanged.
  - Reason's two conditional "Select a reason" `ModelState` errors become `NotNull().When(…)` rules, again first in the list to preserve ordering.
- `IsComplete` and the Reason/CheckAnswers state guards are dropped — the library's path validation already stops users reaching a step they haven't advanced to. Their three redirect tests go with them.
- Test journey seeding moves into a shared `EditMqStatusTestBase`.

## Note

Reason's `UpdateStateAsync` never wrote `AdditionalInformation` / `ProvideAdditionalInformation` to the journey state, so CheckAnswers always read them back as null even though Reason validates them. That's pre-existing behaviour on `main`; I've kept it exactly as-is rather than mixing a behaviour fix into a mechanical migration. Worth a follow-up if it's a real bug.

## Testing

- `just build` — no errors, no new warnings (only the pre-existing `NU1903` transitive-dependency advisories on `main`).
- EditMq/Status tests: **36 passed** (39 before, minus the 3 removed guard tests). All `PageTests.Mqs` + person qualifications tests: **255 passed**.
- Route paths (`/mqs/{qualificationId}/status`, `/status/reason`, `/status/check-answers`) preserved, so the E2E MQ journey tests need no changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/TeachingRecordSystem.SupportUi/Pages/Mqs/EditMq/Status/CheckAnswers.cshtml
+++ b/src/TeachingRecordSystem.SupportUi/Pages/Mqs/EditMq/Status/CheckAnswers.cshtml
@@ -1,11 +1,14 @@
-@page "/mqs/{qualificationId}/status/check-answers/{handler?}"
+@page "/mqs/{qualificationId}/status/check-answers"
 @model TeachingRecordSystem.SupportUi.Pages.Mqs.EditMq.Status.CheckAnswersModel
 @{
     ViewBag.Title = "Check details before editing mandatory qualification";
 }
 
 @section BeforeContent {
-    <govuk-back-link href="@LinkGenerator.Mqs.EditMq.Status.Reason(Model.QualificationId, Model.JourneyInstance!.InstanceId)" />
+    @if (Model.BackLink is not null)
+    {
+        <govuk-back-link href="@Model.BackLink" />
+    }
 }
 
 <div class="govuk-grid-row">
@@ -69,7 +72,7 @@
 
             <div class="govuk-button-group">
                 <govuk-button type="submit">Confirm and update qualification</govuk-button>
-                <govuk-button formaction="@LinkGenerator.Mqs.EditMq.Status.CheckAnswersCancel(Model.QualificationId, Model.JourneyInstance!.InstanceId)" class="govuk-button--secondary" type="submit">Cancel and return to record</govuk-button>
+                <govuk-button name="Cancel" value="true" class="govuk-button--secondary" type="submit">Cancel and return to record</govuk-button>
             </div>
         </form>
     </div>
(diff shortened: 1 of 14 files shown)


gunndabad:
Pushed a follow-up commit fixing the additional-information bug flagged in the PR description, so this branch no longer carries it forward.

**Root cause** was one step deeper than I first described: `EditMqStatusState` had no `ProvideAdditionalInformation` property at all. The Reason page rendered the question and validated the answer, but there was nowhere to put it and the state update never wrote `AdditionalInformation` either — so the user's text was discarded, check answers always showed "None", and the emitted event's `AdditionalInformation` was always null.

**Fix**
- Add `ProvideAdditionalInformation` to `EditMqStatusState`.
- Persist both fields in Reason's `AdvanceTo`, and restore them in `OnGet` so the answers survive navigating back — matching the sibling Reason pages.
- Add the `NotEmpty` "Enter additional detail" rule that every other MQ/alert Reason page has. Without it, answering "Yes" and leaving the box empty passed validation and saved nothing; no existing test posted that combination, so nothing was relying on the old behaviour.

**Tests:** the check answers confirm test now seeds additional information and asserts it reaches the event (it previously asserted `Null` against a state that never set it — green either way). Plus two new tests: the empty-detail validation error, and the GET repopulating from state. EditMq/Status **38 passed** (was 36); all `PageTests.Mqs` + person qualifications **257 passed**. `just build` clean.

hortha (approved):

PR #3643, opened 2026-07-23
Migrate Add Person journey to GOV.UK Questions

gunndabad:
Follows the MQ journeys (#3637–#3642).

## What

Migrates the **Add Person** journey (Index → PersonalDetails → Reason → CheckAnswers) from `FormFlow` to `GovUk.Questions.AspNetCore`.

## Changes

- Add `AddPersonJourneyCoordinator` — `routeValueKeys: []`, as this journey has no scoping route values; drop FormFlow's `IRegisterJourney` from `AddPersonState`.
- **Entry page:** `Index` (`/persons/add`) was outside the journey and redirected to PersonalDetails, which carried `ActivatesJourney`. It's now the `[StartsJourney]` page and `AdvanceTo`s PersonalDetails with `SetAsFirstStep`, matching Add Alert and Add MQ. Cancel still returns to `/persons/add`, which starts a fresh instance exactly as before.
- `CommonJourneyPage` takes the coordinator and exposes a shared `Cancel` bound property and `CancelAsync` helper. The old cancel buttons posted to `?handler=cancel`, which the library rejects as an invalid step — this is the form-field pattern used by the other migrated journeys.
- Back links come from `GetBackLink()`; the "Change" links on check answers use the library's `returnUrl`. `FromCheckAnswers` and the hand-rolled `BackLink` / `NextPage` / `GetPageLink(page, fromCheckAnswers)` chains are gone.
- **Validation → FluentValidation:** PersonalDetails' date-of-birth-in-past, National Insurance number and email `ModelState` errors move into its `InlineValidator`. The parsed NI number and email are exposed as computed properties, so each is parsed once and the same value is reused when writing state (this also avoids CA1806 from a bare `TryParse`). The date rule is written as `!(dateOfBirth > Today)` — the negation of the original — so it behaves identically.
- `IsComplete` / `IsPersonalDetailsComplete` / `IsCreateReasonComplete`, `NextIncompletePage`, `Initialized` / `EnsureInitialized` and the Reason/CheckAnswers guards are dropped — the library's path validation already stops users reaching a step they haven't advanced to. `CommonJourneyPage`'s `OnPageHandlerExecutionAsync` override existed only to call `EnsureInitialized`, so it goes too.
- Test journey seeding moves into a shared `AddPersonTestBase`.

## Test notes

- The three `Get_InvalidState_RedirectsToAppropriatePage` cases go with the guards.
- The two `FromCheckAnswers` tests now pass a real `returnUrl`. Worth knowing: the POST one was *passing* against the unconverted `?fromCheckAnswers=True` — the library treats any unrecognised query parameter as part of the StepId, so the request was an invalid step and got redirected to the furthest seeded step, which happened to be check answers. It now exercises the change-link flow for real.
- The cancel tests posted `cancelButton.FormAction`; they now assert the button is the `Cancel` field and post that field.

## Testing

- `just build` — no errors, no new warnings (only the pre-existing `NU1903` transitive-dependency advisories on `main`).
- AddPerson page tests: **112 passed** (115 before, minus the 3 removed guard cases). All `PageTests.Persons`: **1508 passed**.
- **E2E `AddPersonTests`: 4 passed** — including `AddPerson_NavigateBack` and both check-answers change-link tests, which is the coverage that matters most for the back-link/`returnUrl` rework.
- Route paths (`/persons/add`, `/personal-details`, `/reason`, `/check-answers`) preserved.

One aside: `Persons.IndexTests.Get_WithSearchThatLooksLikeAnEmailAddressWithDifferentCase_DisplaysMatchOnOneLoginUserEmail` fails on a second consecutive run of the suite because it commits a hardcoded email and then self-poisons. It fails the same way on `main` — unrelated to this PR, green on a fresh database.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gunndabad:
Pushed a follow-up commit dropping `CommonJourneyPage`, `AddPersonJourneyPage` and `AddPersonStateBuilder`.

- **`CommonJourneyPage` is gone.** The only genuinely shared behaviour was cancelling — discarding the journey and any evidence uploaded during it — so that moves to `AddPersonJourneyCoordinator.CancelAsync()`, which does the cleanup and returns the URL to redirect to. Each page now derives from `PageModel` directly, injects only what it actually uses, and declares its own `Cancel` bound property (that one can't move — it's a per-page form field).
- **`AddPersonJourneyPage` and `GetPageLink` are gone.** Pages build URLs from `SupportUiLinkGenerator` with `journey.InstanceId`, and the check answers view calls the link generator directly for its "Change" links, matching the alert and MQ check answers pages.
- **`AddPersonStateBuilder` is gone.** All 49 call sites now construct `AddPersonState` with object initialisers through the shared `AddPersonTestBase`, like the other migrated journeys' tests.

One thing worth knowing for future migrations: giving the coordinator constructor dependencies breaks `JourneyHelper`'s default `Activator.CreateInstance` path, so the test base now passes `coordinatorFactory: () => ActivatorUtilities.CreateInstance<AddPersonJourneyCoordinator>(HostFixture.Services)` — same as the Resolve journeys.

I left `AddPersonPostRequestContentBuilder` alone since it builds request content rather than journey state, and the other journeys have no equivalent to compare against — happy to fold it into the shared `FormUrlEncodedContentBuilder`/`MultipartFormDataContentBuilder` helpers if you'd prefer.

**Testing:** `just build` clean (no errors or warnings). AddPerson page tests **112 passed**, E2E `AddPersonTests` **4 passed**.

gunndabad:
Folded in one more commit: removed the dead `MobileNumber` field from `AddPersonPostRequestContentBuilder`.

No page in this journey binds a mobile number, so the field the builder posted was silently ignored — the base `PostRequestContentBuilder` reflects over every public property, so it was going out on the wire and landing nowhere. Dropped the property, its fluent setter and the three call sites, plus the two now-unused `mobileNumber` locals in `CommonPageTests`. 12 lines, no assertions touched it.

The rest of the builder stays as-is: unlike the state builder it isn't duplicating anything an object initialiser does better — the reflection-driven form encoding, `DateOnly` → `.Day`/`.Month`/`.Year` splitting and multipart file parts are real work, and it's one of five such builders sharing a common base.

AddPerson page tests still **112 passed**; `just build` clean.

MrKevJoy (approved):

PR #3685, opened 2026-08-04
Migrate Add Route journey to GOV.UK Questions

gunndabad:
### Context

Continues the move of SupportUi journeys off `WebCommon.FormFlow` and onto
`GovUk.Questions.AspNetCore` (DFE issue #369). This one is Add Route to professional status.
Edit Route and Delete Route are the two left on FormFlow.

Add Route is the first migrated journey whose *shape* depends on its own answers: the route
type and its status decide which of the ten questions get asked. That doesn't fit the
step-at-a-time model the other journeys use, and the difference is most of what's worth
reviewing here.

**Answering a question that changes the journey's shape truncates the path.**

Every question page calls `AddRouteJourneyCoordinator.AnswerAndAdvanceAsync`, which works out
which questions apply either side of the answer:

- **Unchanged** — the normal case, and every question after status: plain `AdvanceTo`. The next
  question is already the next step, so `PushStep` no-ops, the path is left intact, and a
  `returnUrl` pointing at check answers is still in the path, so the library takes the user
  straight back there with its other change links still working.
- **Changed** — only Route and Status can do this: `AdvanceTo` with
  `PushStepOptions { SetAsLastStep = true }`. The steps after this one no longer describe the
  journey, so they're dropped and the user works forward through the questions again.
  `SetAsLastStep` also makes `AdvanceTo` ignore the `returnUrl`, which is what sends them
  forward rather than back to check answers.

The upshot is that the path only ever holds questions the user has actually been through, so
nobody can skip ahead to a question they haven't reached.

That also makes check answers reachable only once every question has been answered, so the old
completeness guard isn't ported: each question's own validation is what stops the journey, and
`AgeRangeFromValidation`/`AgeRangeToValidation` and `EvidenceUploadModelValidator` already cover
the two cases that looked like gaps. The library's `returnUrl` replaces the `fromCheckAnswers`
flag.

**`AddRouteCommonPageModel` and `AddRoutePostStatusPageModel` are gone.** Everything the pages
shared — `PersonId`, `Status`, `GetRouteTypeAsync`, `PageCaption`, `CancelAsync`, advancing,
the unanswered-question lookup — is on the coordinator. Each page is a plain `PageModel`
injecting it. `AddRouteState` is now plain data: its `IsComplete` logic moved to the coordinator
too.

**`PageDriver` is gone.** Its requirement mapping and page ordering moved onto the coordinator as
`GetFieldRequirementForPage` and `GetPagesForRoute`; `AddRoutePage.cs` is now just the enum.
EditRoute's check-answers guard was the only outside caller of the mapping and now calls the
coordinator's static — visibly temporary coupling that goes when that journey is migrated. The
`NextPage`/`PreviousPage` walkers are deleted; back links come from `journey.GetBackLink()`.

**Validation.** Every page validates with FluentValidation, through the new `ThrowIfInvalidAsync`
so model binding errors stop them too — that replaces the hand-rolled `ModelState.AddModelError`
blocks and the `ModelState.IsValid` checks after them. The date inputs carry their error message
prefix on the tag helper rather than a `[DateInput]` attribute.

**Check answers renders its own summary.** The `_RouteDetail` partial and `RouteDetailViewModel`
are deleted and the rows are inline in `CheckAnswers.cshtml`, with the requirement properties from
the shared `RouteDetailModel` moved onto the page model. EditRoute and DeleteRoute keep theirs.

**Guards deleted.** The per-page `BadRequest` checks for an inapplicable route or status
(induction exemption, missing route/status) are unreachable now that path validation runs first.

**Tests.** `AddRouteStateBuilder` is replaced by object initialisers plus a new
`AddRouteTestBase` (`CreateJourneyInstanceAsync`, `CreateChangeReasonDetail`,
`GetJourneyInstanceState`), which seeds the steps of a journey worked through to check answers
via the shared `GetPagesForRoute`. Cancel buttons are form fields now, so those tests POST
`Cancel=true`. The Add Route rows moved out of `EditRoute/PermissionsTests` into their own file,
since that journey can no longer be seeded through FormFlow's `CreateJourneyInstance`.

### Guidance to review

Two changes to test *content* are worth a look rather than a skim:

1. Five theory rows asserted an "(optional)" heading for a question the journey never asks for
   that route and status — holds-from on an *In training* status, for instance. They only
   passed because the old journey had no path validation at all. Each is replaced with a
   `Get_QuestionIsNotAskedForRouteAndStatus_RedirectsToCheckAnswers` test, which pins the
   behaviour that actually matters. A few route/status picks were also random enough to flake
   between 200 and 302 once paths are validated, so they're now filtered.

2. `StatusTests.Post_FromCheckAnswers_WhenNewStatusChangesTheQuestionsAsked_ResumesTheJourneyAtTheNextQuestion`
   is new. It changes the status from check answers so that different questions apply, asserts
   the user is sent forward to the next question rather than back to check answers, and then
   asserts that check answers is no longer reachable until they've worked through them.

3. `Get_IncompleteJourney_RedirectsToExpectedPage` is deleted along with the completeness guard.
   Its 32 rows seeded states the journey can't produce now that the path only holds questions the
   user has been through.

No emitted events change, so `docs/process-type-events.md` is untouched.

Verified locally:

| Suite | Result |
| --- | --- |
| SupportUi.Tests | 3878 passed, 3 skipped |
| SupportUi.EndToEndTests, `AddRouteToProfessionalStatusTests` | 6 passed |
| SupportUi.EndToEndTests (full) | 138 passed, 1 failed |

The E2E failure is `SetStatusTests.Deactivate_CYA_NavigatesBackToCYA`, in a journey this branch
doesn't touch; it passes in isolation and is flaky under parallel load. `just test-changed` also
reports the two AuthorizeAccess `OidcTests` failures that fail on main regardless.

Rebased onto main at `09d68a1bc`, which added `ThrowIfInvalidAsync`; the three conflicting pages
(Status, InductionExemption, Reason) took that call, and the rest of the journey adopted it too.

### Checklist

-   [ ] Attach to Trello card
-   [x] Rebased master


src/TeachingRecordSystem.SupportUi/Pages/RoutesToProfessionalStatus/AddRoute/HoldsFrom.cshtml.cs
```diff
@@ -1,50 +1,57 @@
...
+public class HoldsFromModel(AddRouteJourneyCoordinator journey, TimeProvider timeProvider) : PageModel
 {
+    private readonly InlineValidator<HoldsFromModel> _validator = new()
+    {
+        v => v.RuleFor(m => m.HoldsFrom)
+            .NotNull().WithMessage("Enter the date they first held this professional status")
+            .When(m => m.HoldsFromRequired),
+        v => v.RuleFor(m => m.HoldsFrom)
+            .Must(holdsFrom => !(holdsFrom > timeProvider.Today))
+                .WithMessage("The date they first held this professional status must not be in the future")
```
hortha, line 17: nit: indent

src/TeachingRecordSystem.SupportUi/Pages/RoutesToProfessionalStatus/AddRoute/Reason.cshtml.cs
```diff
@@ -16,7 +17,7 @@ public class ReasonModel(SupportUiLinkGenerator linkGenerator, ReferenceDataCach
             .NotEmpty().WithMessage("Enter a reason")
             .When(m => m.ChangeReason == ChangeReasonOption.AnotherReason),
         v => v.RuleFor(m => m.ProvideAdditionalInformation)
-            .NotNull().WithMessage("Select yes if you want to add more information about why you\u2019re adding this route"),
+            .NotNull().WithMessage("Select yes if you want to add more information about why you’re adding this route"),
         v => v.RuleFor(m => m.AdditionalInformation)
             .MaximumLength(UiDefaults.ReasonDetailsMaxCharacterCount)
                 .WithMessage($"Additional detail {UiDefaults.ReasonDetailsMaxCharacterCountErrorMessage}"),
```
hortha, line 23: nit: indent

hortha (approved):

